PR #48, opened 2015-12-30
Automatically generate pairs config

justmoon:


gip:
Tests: 17 passing but 2 failing on that branch. Do you see the same thing?

gip:
Passing on circle and locally after update.
LGTM

PR #59, opened 2016-01-11
[FEATURE] Support Cases; refactor expiry validation

sentientwaffle:
Requires https://github.com/interledger/five-bells-shared/pull/44

### DIFF
--- a/lib/testPaymentExpiry.js
+++ b/lib/testPaymentExpiry.js
@@ -5,7 +5,7 @@ const request = require('co-request')
 const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
 
 module.exports = function * (config, payment) {
-  let tester = new TransferTester(config, payment)
+  const tester = new TransferTester(config, payment)
   yield tester.loadCaseExpiries()
   return tester
 }
@@ -26,7 +26,7 @@ function TransferTester (config, payment) {
 
 TransferTester.prototype.getExpiry = function (transfer) {
   if (transfer.expires_at) return transfer.expires_at
-  let expiries = _.uniq(_.values(_.pick(
+  const expiries = _.uniq(_.values(_.pick(
     this.expiryByCase, getTransferCases(transfer))))
   if (expiries.length === 0) return undefined
   if (expiries.length === 1) return expiries[0]
@@ -54,7 +54,7 @@ TransferTester.prototype.validateNotExpired = function * () {
 }
 
 TransferTester.prototype.validateTransferNotExpired = function (transfer) {
-  let expires_at = this.getExpiry(transfer)
+  const expires_at = this.getExpiry(transfer)
   if (expires_at && transfer.state !== 'executed' &&
     moment(expires_at, moment.ISO_8601).isBefore(moment())) {
     throw new UnacceptableExpiryError('Transfer has already expired')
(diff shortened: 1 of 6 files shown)


lumberj:
What are cases?

sentientwaffle:
They are what notaries operate on (see https://github.com/interledger/five-bells-notary/blob/master/src/models/case.js)

lib/testPaymentExpiry.js
```diff
@@ -0,0 +1,191 @@
...
+
+module.exports = function * (config, payment) {
+  let tester = new TransferTester(config, payment)
+  yield tester.loadCaseExpiries()
+  return tester
+}
+
+function TransferTester (config, payment) {
+  this.config = config
+  this.source_transfers = payment.source_transfers
```
clark800: Is there a reason for mixing snake case and camel case on the `this` object?

sentientwaffle: no_reasonAt_all, it_is_completelyArbitrary

clark800: It looks like most of the code is using camel case, and that is javascript's built-in case, so I think we should probably stick with that everywhere possible i.e. everywhere that it is not going to a JSON object.

lib/testPaymentExpiry.js
```diff
@@ -0,0 +1,191 @@
+'use strict'
+const _ = require('lodash')
+const moment = require('moment')
+const request = require('co-request')
+const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
+
+module.exports = function * (config, payment) {
+  let tester = new TransferTester(config, payment)
```
clark800: `let` should be avoided when `const` suffices as an assurance against mutation. This is the `prefer-const` rule in eslint. I recommend applying our style guide for better static analysis: https://github.com/ripple/javascript-style-guide

lib/testPaymentExpiry.js
sentientwaffle: Ok, it isn't completely arbitrary. I use snake case where it it is going to a JSON object, or where it is from one (or from the schema). This way, you can search e.g. `/source_transfers`, not `/source_transfers\|sourceTransfers`.

clark800: Yeah, this is a tricky one. I think probably the best option is to have a single module that does camel to snake conversion for all requests/responses so that we can use camel case everywhere else in the code while still only having to search for one thing.

sentientwaffle: The "single module" would have to be reused across fb-{shared,connector,ledger,notary,sender,pathfind,demo} in order to keep the code consistent.

That solution is needlessly complex. The current naming scheme has the substantial perk of mapping directly to -shared's schemas, and requires no additional abstractions. The only downside is one of aesthetics, which is personal preference anyhow.

I think the best option is to leave it as-is.

lumberj: > Ok, it isn't completely arbitrary. I use snake case where it it is going to a JSON object, or where it is from one (or from the schema). This way, you can search e.g. /source_transfers, not /source_transfers\|sourceTransfers.

I like this convention if we follow it strictly. Use snake case for object properties when we're dealing with pure data. Use camelcase for methods and properties of getters/setters on classes.

justmoon:
LGTM

PR #61, opened 2016-01-14
[FEATURE] payment template

sentientwaffle:
Depends on https://github.com/interledger/five-bells-shared/pull/45 and https://github.com/interledger/five-bells-ledger/pull/106

Fixes https://github.com/interledger/five-bells-connector/issues/51
Fixes https://github.com/interledger/five-bells-connector/issues/50

Based on https://github.com/interledger/five-bells-connector/pull/59

PR #63, opened 2016-01-15
[FEATURE] /quote: add liquidity check

### LINKED ISSUE #60: Connector returns error if balance is too low in its account
Add quote error for not enough liquidity in market maker account

sentientwaffle:
Check liquidity before returning a quote. Error if the balance isn't high
enough. Quoting doesn't actually escrow any funds, so no error is not a guarantee
that a transfer preparation will succeed, and an error is not a guarantee that
it wouldn't.

Fixes #60 
Based on https://github.com/interledger/five-bells-connector/pull/61

### DIFF
--- a/controllers/quote.js
+++ b/controllers/quote.js
@@ -1,11 +1,14 @@
 'use strict'
 
 const request = require('co-request')
+const BigNumber = require('bignumber.js')
 const config = require('../services/config')
 const log = require('../services/log')('quote')
 const backend = require('../services/backend')
 const ledgers = require('../services/ledgers')
 const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
+const UnacceptableAmountError = require('../errors/unacceptable-amount-error')
+const ExternalError = require('../errors/external-error')
 
 /* eslint-disable */
 /**
@@ -126,6 +129,12 @@ exports.get = function *() {
   yield query.loadLedgers()
   const quote = yield backend.getQuote(query.toQuoteArgs())
 
+  const sourceBalance = yield getConnectorBalance(query.source_ledger)
+  const sourceAmount = new BigNumber(quote.source_amount)
+  if (sourceBalance.lessThan(sourceAmount)) {
+    throw new UnacceptableAmountError('Insufficient liquidity in market maker account')
+  }
+
   log.debug('' +
     quote.source_amount.toFixed(2) + ' ' +
     query.source_ledger + ' => ' +
@@ -240,7 +249,27 @@ function * getAccountLedger (account) {
   })
   const ledger = res.body && res.body.ledger
   if (res.statusCode !== 200 || !ledger) {
-    throw new Error('Unable to identify ledger from account: ' + account)
+    throw new ExternalError('Unable to identify ledger from account: ' + account)
   }
   return ledger
+}
+
+function * getConnectorBalance (ledger) {
+  const creds = config.ledgerCredentials[ledger]
+  let res
+  try {
+    res = yield request({
+      method: 'get',
+      uri: creds.account_uri,
+      auth: {
+        user: creds.username,
+        pass: creds.password
+      },
+      json: true
+    })
+  } catch (e) { }
+  if (!res || res.statusCode !== 200) {
+    throw new ExternalError('Unable to determine current balance')
+  }
+  return new BigNumber(res.body.balance)
 }
--- a/controllers/quote.js
+++ b/controllers/quote.js
@@ -125,9 +125,8 @@ const ExternalError = require('../errors/external-error')
 /* eslint-enable */
 
 exports.get = function *() {
-  const query = new QuoteQuery(this.query)
-  yield query.loadLedgers()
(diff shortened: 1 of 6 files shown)


lumberj:
> The connector should keep a local cache of its latest balances, based on ledger notifications.

This is the approach I mentioned in the document we've shared regarding the spec

controllers/quote.js
```diff
@@ -121,16 +125,29 @@ const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
...
+  log.debug('' +
+    quote.source_amount.toFixed(2) + ' ' +
+    query.source_ledger + ' => ' +
+    quote.destination_amount.toFixed(2) + ' ' +
+    query.destination_ledger)
+
+  this.body = query.toPaymentTemplate(quote)
+}
 
+function QuoteQuery (params) {
```
lumberj: Why switch to a class from am module in this case? For example, this class is responsible for many things that may not be the responsibility of  `QuoteQuery` object. I'd argue that `QuoteQuery` may be a good way to encapsulate the behavior of parsing and serializing from URI format to JSON, but it should not also be responsible for fetching ledgers, etc.

sentientwaffle: I'll move `toPaymentTemplate` out but I think that the ledger fetching belongs -- it fills in default values for optional parameters.

lumberj: What I mean is that, it doesn't seem to belong as part of the interface of a `QuoteQuery`. It would be more natural to have a function that fetches ledgers, that receives a `QuoteQuery` as an argument.

controllers/quote.js
```diff
@@ -157,37 +173,103 @@ exports.get = function *() {
...
+    throw new Error('Missing required parameter: source_ledger or source_account')
+  }
+  if (this.destination_ledger || this.destination_account) {
+    this.destination_ledger = this.destination_ledger || (yield getAccountLedger(this.destination_account))
+  } else {
+    throw new Error('Missing required parameter: destination_ledger or destination_account')
+  }
+}
+
+QuoteQuery.prototype.toQuoteArgs = function () {
```
lumberj: `toJSON` could be an appropriate and generic method name, as this returns a JSON representation of the object

controllers/quote.js
lumberj: Can we also be consistent and use ES6 classes?

errors/unacceptable-amount-error.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+const UnprocessableEntityError =
+require('five-bells-shared/errors/unprocessable-entity-error')
+
+module.exports = function UnacceptableAmountError (message) {
```
lumberj: Errors are using es6 classes now

lumberj:
LGTM except for a few non-blocker comments on using es6 classes and whether we need to use classes at all for the  quote controller module. Subject can be revisited later.

controllers/quote.js
sentientwaffle: This isn't a generic JSON representation of the `QuoteQuery` -- if it was, it would include expiries, accounts, etc. This is a specific object -- representing arguments for the `getQuote` function, hence `toQuoteArgs`.

controllers/quote.js
```diff
@@ -157,37 +172,103 @@ exports.get = function *() {
...
+  return {
+    source_ledger: query.source_ledger,
+    destination_ledger: query.destination_ledger,
+    source_amount: query.source_amount,
+    destination_amount: query.destination_amount
+  }
+}
+
+function makePaymentTemplate (query, quote) {
+  const source_amount = quote.source_amount.toFixed(2, 2)
```
lumberj: Curious why we're rounding to two decimal places here, and also why we're using different rounding modes for `source_amount` and `destination_amount`

sentientwaffle: I'm not sure about the two decimal places, that's always been there. As for the [rounding modes](https://github.com/MikeMcl/bignumber.js/blob/ad704dc44f9e2de5cd8bf027b0ff3fcc18e83811/bignumber.js#L63-L75), the source amount needs to be rounded up (`CEIL`) -- if it was rounded down, it would cause the transaction to be created at an unacceptable rate, and it would immediately be rejected.

lib/balance-cache.js
```diff
@@ -0,0 +1,47 @@
...
+const BigNumber = require('bignumber.js')
+const ExternalError = require('../errors/external-error')
+
+function BalanceCache (config) {
+  this.config = config
+  this.balanceByLedger = {}
+  this.timer = null
+}
+
+BalanceCache.prototype.get = function * (ledger) {
```
lumberj: Can you add a unit test for the BalanceCache?

controllers/quote.js
lumberj: This is probably a sane default for now, but will probably have to change in the future. The decimal places should probably be based on the precision that the various backends support.

sentientwaffle:
Ok, test added

lumberj:
LGTM

PR #65, opened 2016-01-25
[FIX] create+fund account before subscribing

sentientwaffle:
If the account doesn't exist yet, it can't be subscribed to, so autofund first.

This is necessary to fix the visualization. /cc @MatthewPhinney

### DIFF
--- a/lib/ledgers/five-bells-ledger.js
+++ b/lib/ledgers/five-bells-ledger.js
@@ -91,7 +91,7 @@ FiveBellsLedger.prototype.subscribe = function * (target_uri) {
 }
 
 FiveBellsLedger.prototype._autofund = function * () {
-  log.info('creating account at ' + this.id)
+  log.info('autofunded account at ' + this.id)
   yield request_retry({
     method: 'put',
     url: this.credentials.account_uri,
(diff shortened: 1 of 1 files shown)


lib/ledgers/five-bells-ledger.js
```diff
@@ -87,22 +88,22 @@ FiveBellsLedger.prototype.subscribe = function * (target_uri) {
...
-      body: {
-        name: this.credentials.username,
-        balance: '1500000',
-        connector: config.server.base_uri,
-        password: this.credentials.password
-      }
-    }, 'could not create account at ledger ' + this.id)
-  }
+FiveBellsLedger.prototype._autofund = function * () {
+  log.info('creating account at ' + this.id)
```
clark800: Is it possible that the account already exists?

sentientwaffle: Yes, but it doesn't matter, this is just a debug flag.

clark800: It could cause confusion for the person debugging though

sentientwaffle:
@clark800 ok, I changed the log line

clark800:
LGTM

PR #75, opened 2016-02-08
[FIX] fix condition fulfillment

sentientwaffle:
https://github.com/interledger/five-bells-ledger/pull/133 broke the trader, this should make the compatible again.

### DIFF
--- a/src/lib/executeSourceTransfers.js
+++ b/src/lib/executeSourceTransfers.js
@@ -66,7 +66,10 @@ function * executeSourceTransfers (source_transfers, destination_transfers) {
 
   for (let sourceTransfer of source_transfers) {
     log.debug('requesting fulfillment of source transfer')
-    yield ledgers.putTransfer(sourceTransfer)
+    const transferFulfillment = sourceTransfer.execution_condition_fulfillment
+    if (transferFulfillment) {
+      yield ledgers.putTransferFulfillment(sourceTransfer, transferFulfillment)
+    }
 
     if (sourceTransfer.state !== 'executed') {
       log.error('Attempted to execute source transfer but it was unsucessful')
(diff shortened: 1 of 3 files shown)


src/lib/ledgers/five-bells-ledger.js
```diff
@@ -38,33 +39,48 @@ FiveBellsLedger.prototype.getState = function (transfer) {
...
+    method: 'put',
+    uri: transfer.id,
+    body: transfer
+  })
+  updateTransfer(transfer, updatedTransfer)
+}
+
+FiveBellsLedger.prototype.putTransferFulfillment = function * (transfer, execution_condition_fulfillment) {
+  const updatedTransfer = yield this._request({
+    method: 'put',
```
MatthewPhinney: This should be POST.

src/lib/executeSourceTransfers.js
```diff
@@ -66,7 +66,10 @@ function * executeSourceTransfers (source_transfers, destination_transfers) {
 
   for (let sourceTransfer of source_transfers) {
     log.debug('requesting fulfillment of source transfer')
-    yield ledgers.putTransfer(sourceTransfer)
+    const transferFulfillment = sourceTransfer.execution_condition_fulfillment
```
MatthewPhinney: Is there a separate case for `cancellation_condition_fulfillment`?

src/lib/ledgers/five-bells-ledger.js
clark800: Isn't this POST now?

MatthewPhinney: Discussed with Stefan and we think it is better to use PUT for this. I have a ticket to update the ledger.

src/lib/executeSourceTransfers.js
sentientwaffle: This is for universal mode, so the cancellation condition is not applicable.

MatthewPhinney:
As discussed, we're going to keep fulfillment a put. So this LGTM.

PR #76, opened 2016-02-09
Fix quote rounding logic

emschwartz:
The connector had a bug where it would return a quote that it wouldn't actually accept when submitted as a payment. To avoid this, we should simply round up the source amounts and round down the destination amounts. This will slightly favor the connector (giving them a marginally better rate than their strict minimum), but we can address that later if it is a concern

### DIFF
--- a/src/controllers/quote.js
+++ b/src/controllers/quote.js
@@ -210,8 +210,13 @@ function makeQuoteArgs (query) {
 }
 
 function makePaymentTemplate (query, quote) {
-  const source_amount = quote.source_amount.toFixed(2, 2)
-  const destination_amount = quote.destination_amount.toFixed(2)
+  // The connector will round up the source amounts and round down destination amounts
+  // This fixes a bug where the connector would not accept the quote's values as a payment because of rounding
+  // This rounding method will slightly favor the connector, giving it a marginally better rate than the absolute minimum it would have accepted
+  // TODO it would probably be good to be more precise than 2 decimal places
+  const source_amount = quote.source_amount.toFixed(2, BigNumber.ROUND_UP)
+  const destination_amount = quote.destination_amount.toFixed(2, BigNumber.ROUND_DOWN)
+
   const payment = {
     source_transfers: [{
       type: ledgers.getType(query.source_ledger),


src/controllers/quote.js
```diff
@@ -210,8 +210,9 @@ function makeQuoteArgs (query) {
 }
 
 function makePaymentTemplate (query, quote) {
-  const source_amount = quote.source_amount.toFixed(2, 2)
-  const destination_amount = quote.destination_amount.toFixed(2)
+  // TODO it would probably be good to be more precise than 2 decimal places
+  const source_amount = quote.source_amount.toFixed(2, BigNumber.ROUND_UP)
```
clark800: Maybe add a comment with some of the explanation from the description?

emschwartz: Updated

clark800:
So is this for the quote that the connector is generating? If so, could we do the rounding on the quote rather than the payment template generated from the quote?

emschwartz:
The reason I put it there is that the quote logic is also used to validate the rate for incoming payments (see [here](https://github.com/interledger/five-bells-connector/blob/master/src/lib/payments.js#L244-L257)). Arguably, that piece should use whatever the precise rate is, rather than a rounded value. For the quote API though it makes sense to round to some number of decimal places, because this is what the end user will see. If you disagree with that logic I'm happy to discuss further.

clark800:
It sounds like there are two quotes then: one internal (unrounded) and one external (rounded). It seems like it would be simpler if there was just one quote that applied everywhere.

sentientwaffle:
LGTM. I noticed this bug with the `source_amount`, hence the `quote.source_amount.toFixed(2, 2)`, but it didn't occur to me to round the destination amount down.

emschwartz:
@clark800 I agree that it might be simpler but I don't think it makes sense for the rounding to be done internally. Rounding will lead to the clients getting slightly worse rates than the connector is actually willing to honor. The only reason to do this in the quote is so the amounts returned are relatively human readable. If a connector also happened to expose another API that were not at all intended to be human readable, that might allow clients to get quotes with greater precision. However the client gets the quote, I think the connector should honor the rate it's actually willing to trade at as precisely as possible.

clark800:
@emschwartz Thanks for the explanation. What's the reason for wanting the quotes to be human readable? I imagine human end-users won't be directly interfacing with JSON, so it doesn't seem like a user interface.

emschwartz:
It's a developer interface. If we didn't care about readability at all we'd
have done everything in binary. That might be an option down the line but
it's a lot more painful to develop with when you're iterating a lot

On Fri, Feb 12, 2016 at 10:59 AM, Chris Clark <notifications@github.com>
wrote:

> @emschwartz <https://github.com/emschwartz> Thanks for the explanation.
> What's the reason for wanting the quotes to be human readable? I imagine
> human end-users won't be directly interfacing with JSON, so it doesn't seem
> like a user interface.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/interledger/five-bells-connector/pull/76#issuecomment-183450871>
> .
>



-- 
Evan Schwartz | Software Architect | Ripple
[image: ripple.com] <http://ripple.com>

clark800:
JSON vs binary translation can be done without changing semantics. Rounding is a semantic change, which creates new complexity. I'd recommend against it given the information I have so far, but this is just my two cents, not blocking.

emschwartz:
That's fair. I guess the main issue we're going to run into with these
quotes and rounding is that even if the connector returns a quote with
arbitrarily precise amounts, ledgers will likely not support that degree of
precision. I would actually expect ledgers to define what level of
precision they support, and that'll just have to be something the connector
takes into account to modify it's quotes accordingly.

I don't have a super strong opinion on whether we should make this change
to round consistently or get rid of rounding entirely. However, we do need
to make one of these changes now because at the moment the connector can
return quotes that it won't actually accept when it comes time to make a
payment.

On Fri, Feb 12, 2016 at 11:21 AM, Chris Clark <notifications@github.com>
wrote:

> JSON vs binary translation can be done without changing semantics.
> Rounding is a semantic change, which creates new complexity. I'd recommend
> against it given the information I have so far, but this is just my two
> cents, not blocking.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/interledger/five-bells-connector/pull/76#issuecomment-183456497>
> .
>



-- 
Evan Schwartz | Software Architect | Ripple
[image: ripple.com] <http://ripple.com>

emschwartz:
Replaced by https://github.com/interledger/five-bells-connector/pull/80

PR #78, opened 2016-02-11
[FIX] use co-mocha, so that tests actually run

sentientwaffle:
[Update to use immutable config](https://github.com/interledger/five-bells-connector/pull/70/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L24) included a change that prevented co-mocha from ever executing the tests' generators. As a result, many test failures were going unreported. This remedies that issue, and fixes all of the tests that were broken by that patch and since.

/cc @lumberj

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "start": "node src/index.js",
     "start:watch": "nodemon src/index.js",
     "lint": "eslint .",
-    "test": "istanbul test -- _mocha",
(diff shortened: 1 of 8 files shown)


lumberj:
LGTM

clark800:
In order to allow command line usage of "mocha", you can set the "-r co-mocha" option in the "test/mocha.opts" file.

PR #80, opened 2016-02-16
Fix quote rounding logic and remove fee transfers

emschwartz:
Quote amounts should be rounded in the backend instead of in the controllers. This fixes a bug where connectors would not actually accept quotes they produced.

Fee transfers was an old feature that was never removed. Instead of having fees included as separate transfers in the payment object, they can be paid directly to the connectors with ILP payments (in addition to the spread they charge on FX)


emschwartz:
@clark800 this fixes the quote rounding logic in the manner you suggested

sentientwaffle:
LGTM

src/backends/fixerio/index.js
```diff
@@ -140,6 +140,12 @@ class FixerIoBackend {
         'or destination amount to get quote')
     }
 
+    // Round amounts
+    // TODO Rounding should be based on the level of precision the given ledger supports
+    const AMOUNT_PRECISION = 4
+    sourceAmount = sourceAmount.toFixed(AMOUNT_PRECISION, BigNumber.ROUND_UP)
```
clark800: I try to avoid reassignment like this. There can only be one "sourceAmount" in one context, so either the symbol before the assignment or the symbol after the assignment is "wrong".

Funny story: once I was making this point to another developer and he was like "that's silly, I reassign all the time, that's what variables are for" then a few seconds later we noticed that the reassignment was directly responsible for a bug that we were looking for!

PR #82, opened 2016-02-17
Don't use reassignment in quote rounding

emschwartz:
In response to @clark800 in https://github.com/interledger/five-bells-connector/pull/80#discussion_r53208799


clark800:
LGTM

Issue #79, opened 2016-02-11
five-bells-ledger doesn't send fulfillment on notification anymore


sentientwaffle:
https://github.com/interledger/five-bells-ledger/pull/133 removed the `execution_condition_fulfillment` from the notification. It either needs to be added back, or the notification will have to send the fulfillment instead of the transfer. The latter solution requires changes to the connector.

MatthewPhinney:
My instinct is to keep the fulfillment out of the transfer, since transfers and fulfillments are now persisted in separate db tables in the ledger. We could add a field to the notification body, so it would look something like this:
```
const notificationBody = {
      id: subscriptionURI + '/notifications/' + notification.id,
      subscription: subscriptionURI,
      event: 'transfer.update',
      resource: transfer.getDataExternal(),
      fulfillment: fulfillment.getDataExternal()
    }
```
What do you think @justmoon? This is an API change, so I'd like you to sign off on it.

MatthewPhinney:
I discussed this with Stefan and we decided to go with the notification structure posted above. @sentientwaffle I will ask you to review once I've made the change to the ledger.

MatthewPhinney:
Update: I discussed with Stefan and Evan, and we decided to use the following notification format:
```
{  
   "id":"http://localhost:3004/subscriptions/6edec8b0-e6fd-45da-88b2-11196e7d50b3/notifications/223d7af3-22e8-4dd7-94a2-3c9e06821093",
   "subscription":"http://localhost:3004/subscriptions/6edec8b0-e6fd-45da-88b2-11196e7d50b3",
   "event":"transfer.update",
   "resource":{  
      "debits":[  
         {  
            "amount":"1.00",
            "account":"http://localhost:3004/accounts/millie",
            "authorized":true
         }
      ],
      "credits":[  
         {  
            "account":"http://localhost:3004/accounts/bob",
            "amount":"1.00"
         }
      ],
      "additional_info":{  
         "part_of_payment":"http://localhost:4004/payments/3d928cfb-b71b-4034-a80a-91b8a4b3815b"
      },
      "id":"http://localhost:3004/transfers/e16d4a7d-ad01-43d9-9a85-03673bdf922b",
      "ledger":"http://localhost:3004",
      "state":"executed",
      "expires_at":"2016-02-16T20:40:39.008Z",
      "timeline":{  
         "proposed_at":"2016-02-16T20:40:35.427Z",
[...]

PR #83, opened 2016-02-17
Don't expect fulfillment in transfer

emschwartz:
As of [five-bells-shared v9.2.0](https://github.com/interledger/five-bells-shared/commit/2ebbbbb5e15fd594a7ddd9eea4ffdd99e00f33e4) the `execution_condition_fulfillment` was moved from the transfer object to the notification. This updates the connector accordingly and fixes broken tests.
Fixes https://github.com/interledger/five-bells-connector/issues/79

### DIFF
--- a/src/lib/executeSourceTransfers.js
+++ b/src/lib/executeSourceTransfers.js
@@ -1,76 +1,36 @@
 'use strict'
 
-const _ = require('lodash')
 const ledgers = require('../services/ledgers')
 const log = require('../services/log')('executeSourceTransfers')
-const ExternalError = require('../errors/external-error')
-
-function * addConditionFulfillments (source_transfers, destination_transfers) {
-  for (let sourceTransfer of source_transfers) {
-    // Check if the source transfer's execution_condition is
-    // the execution of the destination transfer
-    let conditionsAreEqual = _.every(destination_transfers,
-      function (destinationTransfer) {
-        return _.isEqual(sourceTransfer.execution_condition,
-          destinationTransfer.execution_condition)
-      })
-
-    if (conditionsAreEqual) {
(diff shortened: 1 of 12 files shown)


test/notificationSpec.js
```diff
@@ -44,6 +44,10 @@ describe('Notifications', function () {
     })
 
     afterEach(function * () {
+      if (nock.pendingMocks() && nock.pendingMocks().length) {
```
sentientwaffle: Why not use an assertion here?

src/lib/executeSourceTransfers.js
```diff
@@ -1,76 +1,36 @@
...
-      if (destinationTransferStateReq.body.message &&
-        destinationTransferStateReq.body.message.state !== 'executed') {
-        log.warn('destination transfer not yet executed')
-      } else {
-        sourceTransfer.execution_condition_fulfillment = {
-          type: destinationTransferStateReq.body.type,
-          signature: destinationTransferStateReq.body.signature
-        }
-      }
+      log.error('Got invalid state receipt: ' + JSON.stringify(stateReceipt))
```
sentientwaffle: If this case is reached, and `undefined` `conditionFulfillment` will be passed to `ledgers.putTransferFulfillment`.

emschwartz: Fixed in https://github.com/interledger/five-bells-connector/commit/bf1a54ce72f5b8ff140879d39e806721c33e1fa6

test/notificationSpec.js
emschwartz: Fixed in https://github.com/interledger/five-bells-connector/commit/4d75c2d1785199f9b3793314a2cba4a27ce47ef4

justmoon:
LGTM

PR #87, opened 2016-02-22
Enable using client certificate authentication

lumberj:


lumberj:
cc: @geertweening @justmoon @gip

sentientwaffle:
LGTM

src/lib/config.js
```diff
@@ -11,13 +13,65 @@ function isRunningTests () {
...
 }
 
+function parseCredentialsEnv () {
+  return JSON.parse(Config.getEnv(envPrefix, 'CREDENTIALS') || '{}')
+}
+
+function parseCredentials () {
+  const credentialsEnv = parseCredentialsEnv()
+
+  return _.reduce(credentialsEnv, (parsed, credentials, ledger) => {
```
clark800: `keyBy` and `map` might be simpler than `reduce`

src/lib/config.js
```diff
@@ -11,13 +13,65 @@ function isRunningTests () {
...
+
+    return parsed
+  }, {})
+}
+
+function validateLocalEnvConfig () {
+  const credentials = parseCredentialsEnv()
+
+  _.forEach(credentials, (credential, ledger) => {
+    if (credential.username === undefined ^ credential.password === undefined) {
```
clark800: This does work, but `^` is the bitwise XOR operator, not a boolean XOR operator, which means this returns `0` or `1`, which would be a type error if we apply type checking

clark800: Note: boolean XOR operator is `!==`

test/mocha.opts
```diff
@@ -0,0 +1 @@
+-r co-mocha
```
clark800: With this I think we can remove the option from package.json

src/lib/config.js
lumberj: Sure, I can change this to 

```js
credential.username === undefined !== credential.password === undefined
```

clark800: don't forget parens :)

src/lib/config.js
lumberj: I can refactor this in the future, as it would require a major version bump for lodash

test/mocha.opts
lumberj: Removed

PR #88, opened 2016-02-23
Update five-bells-shared to 10.0.0

lumberj:


lumberj:
The breaking change from v9 to 10 is here, https://github.com/interledger/five-bells-shared/pull/89

MatthewPhinney:
LGTM

PR #89, opened 2016-02-23
Ilp-quoter

gip:


emschwartz:
Could you rebase onto the latest master? It looks like your second commit is merging master into your branch rather than the other way around

PR #90, opened 2016-02-23
Ilp quoter backend

gip:


src/backends/ilp-quoter/index.js
```diff
@@ -0,0 +1,105 @@
...
+      destination_ledger: params.destination_ledger,
+      source_amount: result.body.source_amount,
+      destination_amount: result.body.destination_amount
+    }
+    return quote
+  }
+
+  * hasPair (source, destination) {
+    const pair = lookupCurrencies(source, destination)
+    const r = this.pairs.find((p) => (p[0] === pair[0]) && (p[1] === pair[1])) !== undefined
```
MatthewPhinney: Can you use a more descriptive variable name (instead of `r`)?

src/backends/ilp-quoter/index.js
```diff
@@ -0,0 +1,105 @@
...
+const NoAmountSpecifiedError = 
+  require('../../errors/no-amount-specified-error')
+const UnacceptableQuoterAmountError = 
+  require('../../errors/unacceptable-quoter-amount-error')
+const log = require('../../services/log')('ilpquote')
+const config = require('../../services/config')
+
+function lookupCurrencies (source_ledger, destination_ledger) {
+  for (let pair of config.get('tradingPairs').toJS()) {
+    if (pair[0].indexOf(source_ledger) === 4 &&
```
MatthewPhinney: What is the significance of `4`?

src/backends/ilp-quoter/index.js
```diff
@@ -0,0 +1,105 @@
...
+  return null
+}
+
+/**
+ * Example backend that connects to an external component to get
+ *   the source and destination amounts
+ */
+class ILPQuoter {
+  constructor (opts) {
+    log.warn('ctor')
```
MatthewPhinney: Shouldn't this be a `debug` log?

src/backends/ilp-quoter/index.js
gip: This code was copied from the fixerio backend. I'll look into it.

emschwartz:
Could you include more of a description in the commit for what this functionality is for?

src/backends/ilp-quoter/index.js
```diff
@@ -0,0 +1,104 @@
...
+                '?type=' + type
+    const result = yield request({ uri, json: true })
+    const fixedAmount = type === 'source' ? result.body.source_amount
+                                          : result.body.destination_amount
+    if (fixedAmount !== amount.toString()) {
+      throw new UnacceptableQuoterAmountError('Backend returned an invalid ' +
+                type + ' amount: ' + fixedAmount + ' (expected: ' + amount + ')')
+    }
+    const quote = {
+      source_ledger: params.source_ledger,
```
emschwartz: It should be documented somewhere that the amounts should be queried and returned as strings

src/backends/ilp-quoter/index.js
```diff
@@ -0,0 +1,104 @@
...
+      destination_ledger: params.destination_ledger,
+      source_amount: result.body.source_amount,
+      destination_amount: result.body.destination_amount
```
emschwartz: It should be documented somewhere that the amounts should be queried and returned as strings

gip: Yes - the ilp-quoter backend doesn't do any arithmetic -- it's up to the external component to compute the correct amounts. So strings are used at this point. Where are the docs for the connector component located?

emschwartz: General documentation is lacking right now, but some comments in the code should do it for now

gip: Basic docs added.

src/backends/ilp-quoter/index.js
MatthewPhinney: Maybe add a comment to the code for clarity? I find it hard to follow without the explanation. Also, I would argue for naming the currency variables:
```
const currencyA = pair[0].slice(0, 3)
const currencyB = pair[1].slice(0, 3)
return [currencyA, currencyB]
```
or something like that. Otherwise it's not clear to me what this method returns.

gip: Sure. Done.

MatthewPhinney:
LGTM

src/backends/ilp-quoter/index.js
clark800: I think it would be more clear to do something like `pair.split('@', 2)[1]` and have a separate method that parses the string format into a POJO along with a comment that shows an example like `USD@http://localhost:3000`

gip: Sure - pull request welcome. Note that I expect the way to configure the (currency, ledger) to change in the near future so this code will be refactored then anyway.

PR #96, opened 2016-02-29
Improved handshake with backend

gip:


lumberj:
Can you add a test for this?

gip:
Sure - though that will require a redesign of the way to handle configuration in the connector. So that will take some extra time.

lumberj:
I think the investment to do that, as discussed in person, is worthwhile because it will reduce the probability of regressions and it is a much smaller investment in time today than it will be in a few months.

gip:
I totally agree, I was just updating the ticket to capture our discussion :)

lumberj:
LGTM -- discussed offline, will followup with unit tests to unblock integration testing for now

PR #97, opened 2016-03-01
Allow using TLS cert credentials for admin account (autofund feature)

lumberj:
Note: autofunding only used for testing purposes

### DIFF
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -5,7 +5,7 @@ const Config = require('five-bells-shared').Config
 const Utils = require('../lib/utils')
 const _ = require('lodash')
 
-const envPrefix = 'connector'
+const envPrefix = 'CONNECTOR'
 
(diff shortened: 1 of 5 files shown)


src/lib/config.js
```diff
@@ -47,6 +47,22 @@ function parseCredentials () {
   }, {})
 }
 
+function parseAdminEnv () {
+  const adminUser = Config.getEnv(envPrefix, 'ADMIN_USER') || 'admin'
```
gip: Why do we need to default to `admin`? A real, non-default user name should always be specified I'd think.

gip: LGTM

lumberj: This is existing behavior

lumberj: Since this is only for debug / testing, it shouldn't matter

PR #98, opened 2016-03-01
Add `npm version` pre and post tasks for release

lumberj:
Use `npm version {patch|minor|major}` to release new version

https://docs.npmjs.com/cli/version

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,9 @@
     "test": "istanbul test -- _mocha",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/.bin/coveralls",
     "apidoc": "apidoc -o apidoc-out -i src/controllers/",
-    "integration": "integration all"
+    "integration": "integration all",
+    "preversion": "npm test",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,7 +24,9 @@
     "test": "istanbul test -- _mocha",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/.bin/coveralls",
     "apidoc": "apidoc -o apidoc-out -i src/controllers/",
-    "integration": "integration all"
+    "integration": "integration all",
+    "preversion": "npm test",
+    "postversion": "git push --follow-tags"
```
clark800: Is it guaranteed to be published before the tag is pushed?

lumberj: Do you mean that `npm version` should also run `npm publish`? It does not by default

lumberj: The publish step is done by CI, so it's guaranteed that the tag is pushed first

gip:
LGTM. Very cool feature.

PR #99, opened 2016-03-02
Use a script to bump a version and push to master

lumberj:
Adds `npm run bump` which will:

1) Check we're on master
2) Check local is in sync with remote (not diverged, ahead or behind)
2a) Print a list of the commits since the last released version
3) Bump version locally, commit and tag
4) Push commit and tags to remote

### DIFF
--- a/bin/bump.sh
+++ b/bin/bump.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+
+# http://www.tldp.org/LDP/abs/html/options.html
 set -eo pipefail
 
 # http://stackoverflow.com/questions/1593051/how-to-programmatically-determine-the-current-checked-out-git-branch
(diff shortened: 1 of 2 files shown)


lumberj:
cc @geertweening

geertweening:
This looks good, +1 would like someone else to review as well

lumberj:
@sentientwaffle reviewed this for correctness, and I also used [shellcheck](www.shellcheck.net)

bin/bump.sh
```diff
@@ -0,0 +1,90 @@
...
+      npm version major
+      ;;
+    minor)
+      npm version minor
+      ;;
+    patch)
+      npm version patch
+      ;;
+    *)
+      echo "Usage npm run bump {major|minor|patch}"
```
clark800: Did you test this usage? The npm docs say that you are supposed to use `--`: https://docs.npmjs.com/cli/run-script

lumberj: Yes, works as expected

PR #101, opened 2016-03-02
FIX: Pass username into client-cert credentials

lumberj:
Also, add fingerprint when autofunding an account (debug feature) so it works
with client cert auth


MatthewPhinney:
LGTM (once tests pass).

PR #104, opened 2016-03-04
[FEATURE] Don't trust transfers from the sender

### LINKED ISSUE #94: Connector authorizes debig although transfer object is in proposed state
In the error case reported in RILP-254, Connector sets 'authorized' flag in debit to 'true', even though the transfer object is in 'proposed' state (i.e., not yet in 'prepared' state).  
It is set in settlt() function in src/lib/payments.js. 
I have talked with @justmoon and he thinks Connector should not do this. 

[connector-payment.json.txt](https://github.com/interledger/five-bells-connector/files/149249/connector-payment.json.txt)
[connector-receiver-transfer.json.txt](https://github.com/interledger/five-bells-connector/files/149250/connector-receiver-transfer.json.txt)
[connector-sender-transfer.json.txt](https://github.com/interledger/five-bells-connector/files/149248/connector-sender-transfer.json.txt)

sentientwaffle:
Fixes #94


gip:
If the connector instance crashes, data in settlementQueue will be lost. What is the recovery mechanism -- is the transfer simply canceled and the sender needs to start another transfer?

sentientwaffle:
There's a TODO in `src/lib/settlementQueue.js` suggesting that this be moved the the db eventually. Right now if the connector crashes, the destination transfers will never get authorized, so the payment will eventually expire.

gip:
I checked and the PR passes our integration tests. LGTM.

src/controllers/payments.js
```diff
@@ -140,9 +141,11 @@ exports.put = function *(id) {
 
   payment.id = id.toLowerCase()
 
-  let isPrepared = yield Payments.validate(payment)
-  if (isPrepared) {
-    yield Payments.settle(payment)
+  yield Payments.validate(payment)
+  const truePayment = settlementQueue.storePayment(payment)
```
clark800: The name `truePayment` doesn't really give me any idea how it differs from `payment`, is there another name that describes the difference between the two?

sentientwaffle: How about `trustedPayment`

clark800: Sounds better, how does it become trusted? Once it goes in the queue it becomes trusted?

sentientwaffle: The comment in `settlementQueue.js` explains it.

src/lib/settlementQueue.js
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Called via POST /notifications (from a ledger)
+ *
+ * @param {Transfer} transfer a trusted source_transfer
+ * @returns {Payment|undefined} the trusted payment
+ */
+SettlementQueue.prototype.storeTransfer = function (transfer) {
+  if (transfer.state !== 'prepared' && transfer.state !== 'executed') {
```
naoitoi: For state strings, please use constant values defined in shared. 
Example: https://github.com/interledger/five-bells-ledger/commit/af4831d3507712baf5abaf3edad1390a3439f2f1

src/controllers/payments.js
clark800: Ok, I think I understand roughly what's going on. It's a bit confusing that `storePayment` returns a new type of "payment". Maybe it would be more clear to split this up into two function calls, one that just stores the payment and returns nothing and one that gets the trusted payment.

src/lib/settlementQueue.js
sentientwaffle: Connector doesn't use those anywhere else. If someone wants to refactor that eventually, they can change them all at once.

src/controllers/payments.js
sentientwaffle: The operations belong together. I agree that `store` isn't the most descriptive verb for this, but the doc and code make it clear whats going on, so I don't think its an issue.

src/lib/settlementQueue.js
naoitoi: Sure, someone could do it in refactoring, but the purpose of putting the constants in Shared was to reduce the risk of mistyping state strings.  So - I think it's reasonable for new code to start using the constants.  What do you think?

naoitoi:
This is more of a question than a review comment: 
Is a notification message from Ledger verified by Connector (via digital signature or something similar)?  
In other words, now Connector relies on what comes from Ledger.  What mechanism do we use to ensure this message indeed came from Ledger? 
Perhaps this is already implemented, but I'd appreciate it if you could point me to it.

sentientwaffle:
That isn't implemented yet.

src/lib/settlementQueue.js
sentientwaffle: I think that a combination of autocomplete and testing means that a typo in a state string isn't a significant risk, and the constants just add unnecessary complexity. Regardless, they should either be used everywhere or nowhere, and since this isn't a refactoring PR I'll just keep using the literal strings.

naoitoi:
Thank you, DJ.  Wouldn't the lack of verifying Ledger's notification present a similar risk to what this ticket is for?  
In addition to sending a faked transfer to Connector, can the attacker send a faked Ledger notification to Connector, and achieve the same result?

sentientwaffle:
Yes, but trusting transfer states on `/payments/...` is a bug (that this fixes), whereas verifying notifications is a feature that hasn't been implemented yet.

naoitoi:
LGTM.

PR #106, opened 2016-03-08
FIX: Enable using client-cert auth on unsubscribe request

lumberj:


src/lib/ledgers/five-bells-ledger.js
```diff
@@ -168,10 +162,10 @@ function * request_retry (opts, error_msg) {
 }
 
 FiveBellsLedger.prototype.unsubscribe = function * () {
-  yield request({
+  yield request_retry({
     method: 'delete',
     url: this.id + '/subscriptions/' + notificationUuid
-  })
+  }, this.credentials)
```
sentientwaffle: `request_retry` takes an `error_msg` parameter here.

src/lib/ledgers/five-bells-ledger.js
```diff
@@ -146,18 +134,24 @@ FiveBellsLedger.prototype._autofund = function * () {
...
 }
 
-function * request_retry (opts, error_msg) {
+function * request_retry (opts, error_msg, credentials) {
   let delay = backoffMin
   while (true) {
     try {
-      let res = yield request(opts)
+      let res = yield request(lodash.defaults(opts, {
+        auth: credentials && {
```
sentientwaffle: Is it `credentials.username/password` or `credentials.user/pass`? Pick one. Also, instead of `credentials && {...` it should be `credentials.password && {...` so that we don't send basic auth of undefined:undefined when using client cert auth.

lumberj: This was like this before I made this change:

For admin it's user/pass
https://github.com/interledger/five-bells-connector/blob/master/src/lib/config.js#L58

For ledger credentials it's username/password
https://github.com/interledger/five-bells-connector/#trading

How is it possible to send basic-auth of undefined:undefined...?

lumberj: I can change the admin credentials object to match username/password in this PR. It should be a small refactor

lumberj: In the config we take care of the case missing username/password/key, https://github.com/interledger/five-bells-connector/blob/1c58fa9d99c0e57d6fcfb4fa0eefa3b165233534/src/lib/config.js#L123-L128

lumberj: Updated

sentientwaffle: Oops, you're right about user/pass/username/password.

When credentials is doesn't include the basic auth credentials (e.g. if its using client certs instead) an auth object of `{user: undefined, pass: undefined}` will still get passed to request.

lumberj: You're right

lumberj: Fixed that

sentientwaffle:
LGTM

PR #109, opened 2016-03-11
[BREAKING] Send transfers in the memo

sentientwaffle:
Depends on https://github.com/interledger/five-bells-shared/pull/111

Spec/background: https://docs.google.com/document/d/1zTLQLBx0p6-Ds9Ki8nCaVBdXUK4mbJZDQOaz83n6VY8/edit

/cc @justmoon


emschwartz:
It looks like the connector isn't putting the `source_transfer` in the `destination_transfer`'s memo. Isn't that also part of the point? Without the payments functionality, how does the connector know which source transfer to execute when it gets an incoming notification for an executed `destination_transfer`

emschwartz:
I wonder if it would actually make sense to keep the name `source_transfer` and `destination_transfer`, rather than switch to `next` and `previous`

sentientwaffle:
Well, the connector only supports one-to-one payments now, so it is using a mapping in `destinationSubscriptions`. Once the new fulfillments API is implemented, though, it will just need a `source_ledger` in the memo.

And I like that idea re:memo naming.

emschwartz:
Actually, right now all you really need is the source transfer id to know where to send the fulfillment. I think you could get rid of the `destinationSubscriptions` now and just include the id in the destination transfer debit memo. Then, when the new fulfillment API is ready you just change it to be the fulfillment endpoint instead of the transfer ID.

src/lib/payments.js
```diff
@@ -357,17 +353,29 @@ Payments.prototype.updateTransfer = function * (updatedTransfer, relatedResource
...
     'payment we have a record of or the corresponding source ' +
     'transfers may already have been executed')
 }
 
+Payments.prototype.updateSourceTransfer = function * (updatedTransfer, traderCredit) {
+  const destinationTransfer = traderCredit.memo
+  if (!destinationTransfer) return
+
+  // The source transfer isn't prepared yet, so ignore it.
+  if (updatedTransfer.state !== 'prepared' && updatedTransfer.state !== 'executed') {
```
lumberj: Can use a explaining variable here instead of a comment

```js
const isTransferReady = updatedTransfer.state === 'prepared' || updatedTransfer.state === 'executed
if (!isTransferReady) {
..
}
```

test/notificationSpec.js
```diff
@@ -161,13 +155,14 @@ describe('Notifications', function () {
...
-        .expect(201)
-        .end()
-      yield this.request()
         .post('/notifications')
-        .send(this.notificationSourceTransferPrepared)
+        .send({
+          id: this.notificationSourceTransferPrepared.id,
+          event: 'transfer.update',
+          resource: _.merge({}, payment.source_transfers[0], {
+            credits: [{memo: payment.destination_transfers[0]}]
```
lumberj: What if the memo is not a properly formatted transfer?

emschwartz:
@justmoon pointed out that some of the functionality tested in the tests that are being removed by this PR should probably be moved over into the notifications tests

clark800:
Nice to see lots of red lines!

justmoon:
Agree. That said, some of the red lines are tests...

@sentientwaffle Please make sure that we add the new equivalent of the old tests. A lot of the things that we previously tested by PUTting to /payments now has to be tested by pretending to submit a ledger notification with a memo.

We need to make sure we don't massively drop our test coverage before we merge this.

sentientwaffle:
Ok, I added back the relevant tests.

src/controllers/payments.js
```diff
@@ -1,156 +0,0 @@
...
- *           "message_hash": "claZQU7qkFz7smkAVtQp9ekUCc5LgoeN9W3RItIzykNEDbGSvzeHvOk9v/vrPpm+XWx5VFjd/sVbM2SLnCpxLw==",
- *           "signer": "http://ledger.example",
- *           "type": "ed25519-sha512",
- *           "public_key": "Lvf3YtnHLMER+VHT0aaeEJF+7WQcvp4iKZAdvMVto7c="
- *         },
- *         "expires_at": "2015-06-16T00:00:10.000Z",
- *         "state": "proposed"
- *       }]
- *    }'
- *    https://connector.example/payments/c9377529-d7df-4aa1-ae37-ad5148612003
```
emschwartz: Since this is coming out of the apidocs we probably want to add documentation to the readme of how this works now

src/lib/executeSourceTransfers.js
```diff
@@ -58,15 +59,20 @@ function * executeSourceTransfers (sourceTransfers, destinationTransfers, relate
...
+
+    log.debug('Requesting fulfillment of source transfer: ' + sourceTransferID + ' (fulfillment: ' + JSON.stringify(conditionFulfillment) + ')')
     // TODO check the timestamp on the response from the ledger against
     // the transfer's expiry date
     // See https://github.com/interledger/five-bells-ledger/issues/149
-    yield ledgers.putTransferFulfillment(sourceTransfer, conditionFulfillment)
-    if (sourceTransfer.state !== 'executed') {
-      log.error('Attempted to execute source transfer but it was unsucessful')
-      log.debug(sourceTransfer)
+    const sourceTransferState = yield ledgers.putTransferFulfillment(sourceTransferLedger, sourceTransferID, conditionFulfillment)
```
emschwartz: This looks like it doesn't match the function signature of `putTransferFulfillment`, which expects just `transferID, execution_condition_fulfillment`

test/data/notificationSourceTransferAtomic_TwoCases.json
```diff
@@ -0,0 +1,72 @@
...
+              "source_transfer_ledger": "http://usd-ledger.example/USD",
+              "source_transfer_id": "http://usd-ledger.example/USD/transfers/6851929f-5a91-4d02-b9f4-4ae6b7f1768c"
+            }
+          }],
+          "credits":[{
+            "amount":"1.00",
+            "account":"http://eur-ledger.example/accounts/bob"
+          }],
+          "execution_condition": {
+            "message_hash": "9EXcGgJa0zTIUxEQXaMKpvoGsNmYSxv8J6C12LBNA5hRMfmbwDCzAU4O14UKnqcIaQqMGULHLI+T6Nk3+qYd3Q==",
```
emschwartz: We could definitely use a more compressed format for including the next transfer details in the memo because a lot of this information can be inferred instead of duplicated. But that's for another PR

emschwartz:
LGTM

PR #110, opened 2016-03-11
Create domain models

lumberj:
First step for creating a layered architecture for connector
https://github.com/interledger/five-bells-ledger/issues/112


src/models/metadata.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+const config = require('../services/config')
+
+function getMetadata () {
+  return {
+    public_key: config.getIn(['keys', 'ed25519', 'public']),
+    urls: {
```
clark800: The "urls" object should probably be added by the controller since it's an HTTP thing. And for clarity this method could be renamed to something like "getDomainMetadata", which then gets merged with HTTP metadata.

lumberj: In this PR, I only moved as much of the existing code as I could without doing any refactoring. I plan to do anything else in a followup.

src/models/payments.js
```diff
@@ -0,0 +1,22 @@
...
+  if (trustedPayment) {
+    yield paymentsService.settle(trustedPayment)
+    return trustedPayment
+  }
+
+  return payment
+}
+
+module.exports = {
+  create: create
```
clark800: To be consistent with other methods, this should probably be `createPayment`.

src/models/metadata.js
lumberj: There are some places where it's more difficult to separate the HTTP concerns from the domain layer

clark800: Well this whole pull request is refactoring. It's a little strange to move something if you intend to move it right back.

clark800:
LGTM after rename of `createPayment`

sentientwaffle:
Some of the payment stuff is getting removed by https://github.com/interledger/five-bells-connector/pull/109 anyway (they're going to conflict).

lumberj:
@sentientwaffle I'll wait until that is merged

sentientwaffle:
@lumberj thanks!

lumberj:
@clark800 updated

PR #113, opened 2016-03-18
Change default precision in Fixer backend

lumberj:
Default scale (digits after decimal point) in the ledger is 2

Requires https://github.com/interledger/five-bells-integration-test/pull/13

### DIFF
--- a/src/backends/fixerio/index.js
+++ b/src/backends/fixerio/index.js
@@ -142,7 +142,7 @@ class FixerIoBackend {
 
     // Round amounts
     // TODO Rounding should be based on the level of precision the given ledger supports
-    const AMOUNT_PRECISION = 4
+    const AMOUNT_PRECISION = 2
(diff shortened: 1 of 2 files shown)


src/backends/fixerio/index.js
```diff
@@ -142,7 +142,7 @@ class FixerIoBackend {
 
     // Round amounts
     // TODO Rounding should be based on the level of precision the given ledger supports
-    const AMOUNT_PRECISION = 4
+    const AMOUNT_PRECISION = 2
```
emschwartz: Can we get the precision from the ledger instead of having it fixed?

lumberj: Yes, I'm working on that next!

lumberj: Related: https://github.com/interledger/five-bells-ledger/pull/204

PR #116, opened 2016-03-21
Create payments domain model

lumberj:
Remove services/payments and services/subscriptions
Use a functional interface

This is part of refactoring to a three-layer architecture


src/models/payments.js
```diff
@@ -0,0 +1,376 @@
+'use strict'
```
lumberj: I wonder if this module should really be called `transfers`

lumberj:
cc @gip

src/models/subscriptions.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
+
+const _ = require('lodash')
+const log = require('../services/log')('subscriptions')
+const payments = require('../models/payments')
+
+function * subscribePairs (pairs, ledgersService, config) {
+  let ledgers = _(pairs)
```
clark800: const

clark800:
LGTM

PR #118, opened 2016-03-22
Fix exception on payment that is missing the execution_condition

clark800:
Addresses RILP-362. I'm not sure how to test this now since the payments endpoint was removed.

### DIFF
--- a/src/models/payments.js
+++ b/src/models/payments.js
@@ -28,6 +28,10 @@ function sourceConditionIsDestinationTransfer (source, destination) {
     state: 'executed'
   }
 
+  if (!source.execution_condition) {
+    return false
+  }
+
   if (source.execution_condition.message_hash &&
     source.execution_condition.message_hash !== hashJSON(expectedMessage)) {
     return false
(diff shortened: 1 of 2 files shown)


lumberj:
@clark800 looks like all the tests for these conditions are in https://github.com/interledger/five-bells-connector/blob/master/test/notificationSpec.js

emschwartz:
This fix is technically correct because it's in the function `sourceConditionIsDestinationTransfer`. However, we should actually allow payments where the source transfer has no execution condition (irrespective of whether or not the destination transfer does). That should be handled as a separate check though

lumberj:
@clark800 is this already covered by one of the test cases in notificationSpec? Otherwise LGTM

clark800:
This isn't covered because the tests pass with or without this change. It is not clear to me how to add a test case.

lumberj:
@clark800 you can modify this test case, and write a failing test. https://github.com/interledger/five-bells-connector/blob/master/test/notificationSpec.js#L199 -- the connector receives a notification with a payment in the memo missing an execution condition.

clark800:
@lumberj updated

lumberj:
LGTM

PR #121, opened 2016-03-23
Scalability fix.  Do not respond to Ledger with 422 for a non-critica…

naoitoi:
…l error.

### DIFF
--- a/src/controllers/notifications.js
+++ b/src/controllers/notifications.js
@@ -3,6 +3,7 @@
 const requestUtil = require('five-bells-shared/utils/request')
 const log = require('../services/log')('notifications')
 const model = require('../models/notifications')
+const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
 
 /* eslint-disable */
 /**
@@ -70,7 +71,19 @@ const model = require('../models/notifications')
 exports.post = function * postNotification () {
   const notification = yield requestUtil.validateBody(this, 'Notification')
   log.debug('Got notification: ' + JSON.stringify(notification))
-  yield model.processNotification(notification)
-
+  try {
+    yield model.processNotification(notification)
+  } catch (e) {
+    if (!(e instanceof UnacceptableExpiryError)) {
+      log.error('Notification handling received critical error: ' + e)
+      throw e
+    } else {
+      // If this receives UnacceptableExpiryError, do not respond with HTTP error 422.
+      // Doing so would cause Ledger to reply with the same message, creating an infinite loop.
+      // Error handling mechanism for mode.processNotifiation() should be improved in the future.
+      // It should not throw an error in this situation.
+      log.debug('Notification handling received non-critical error: ' + e)
+    }
+  }
   this.status = 200
 }
--- a/src/lib/testPaymentExpiry.js
+++ b/src/lib/testPaymentExpiry.js
@@ -176,7 +176,7 @@ TransferTester.prototype.loadCaseExpiry = function * (caseID) {
     json: true
   })
   if (caseRes.statusCode !== 200) {
-    throw new Error('Unexpected remote error: ' + caseRes.statusCode + ' ' + caseRes.body)
+    throw new UnacceptableExpiryError('Unexpected remote error: ' + caseRes.statusCode + ' ' + caseRes.body)
   }
   const expiry = caseRes.body.expires_at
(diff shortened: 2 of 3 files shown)


src/controllers/notifications.js
```diff
@@ -70,7 +71,19 @@ const model = require('../models/notifications')
...
-  yield model.processNotification(notification)
-
+  try {
+    yield model.processNotification(notification)
+  } catch (e) {
+    // If this receives UnacceptableExpiryError, do not respond with HTML error 422.
+    // Doing so would cause Ledger to reply with the same message, creating an infinite loop.
+    if (e instanceof UnacceptableExpiryError) {
+      log.debug('Notification handling received non-critical error: ' + e)
+      this.status = 202
```
MatthewPhinney: What was the thinking behind using `202` here?

MatthewPhinney:
LGTM

src/controllers/notifications.js
lumberj: Same question

src/controllers/notifications.js
```diff
@@ -70,7 +71,19 @@ const model = require('../models/notifications')
...
   const notification = yield requestUtil.validateBody(this, 'Notification')
   log.debug('Got notification: ' + JSON.stringify(notification))
-  yield model.processNotification(notification)
-
+  try {
+    yield model.processNotification(notification)
+  } catch (e) {
+    // If this receives UnacceptableExpiryError, do not respond with HTML error 422.
+    // Doing so would cause Ledger to reply with the same message, creating an infinite loop.
+    if (e instanceof UnacceptableExpiryError) {
```
lumberj: Why is the model throwing an exception if it should be ignored? It probably makes sense to put this business logic into the model

src/controllers/notifications.js
naoitoi: As discussed with Matt offline, I picked 202 because I'd like Connector to return success status code (2xx), because otherwise Ledger would resend notifications.  Within all 2xx's, this one (202 - Accepted) seems to be a good semantic match, as Connector accepted the notification, even though it didn't act on it.

src/controllers/notifications.js
naoitoi: @lumberj I agree.  The model shouldn't be throwing an exception for situations like this.  It should be using a different way of error handling mechanism.  The problem is that it relies on throwing an error currently, and fixing that would be a refactoring job that's out of this PR's scope.

lumberj: I think the problem is that there is not a clear separation of concerns between whats in the models/payments models/notifications. The logic for checking expiry should be in models/notifications. Maybe you can add the `catch` in models/notifications for now, and it can return a value that can be used to determine the status code.

naoitoi: As discussed with Matt offline, I picked 202 because I'd like Connector to return success status code (2xx), because otherwise Ledger would resend notifications. Within all 2xx's, this one (202 - Accepted) seems to be a good semantic match, as Connector accepted the notification, even though it didn't act on it.

src/controllers/notifications.js
naoitoi: As discussed with Matt offline, I picked 202 because I'd like Connector to return success status code (2xx), because otherwise Ledger would resend notifications. Within all 2xx's, this one (202 - Accepted) seems to be a good semantic match, as Connector accepted the notification, even though it didn't act on it.

src/controllers/notifications.js
lumberj: Add try/catchhere, https://github.com/interledger/five-bells-connector/blob/master/src/models/notifications.js#L7

src/controllers/notifications.js
```diff
@@ -70,7 +71,19 @@ const model = require('../models/notifications')
...
-  yield model.processNotification(notification)
-
+  try {
+    yield model.processNotification(notification)
+  } catch (e) {
+    if (!(e instanceof UnacceptableExpiryError)) {
+      log.error('Notification handling received critical error: ' + e)
+      throw e
+    } else {
+      // If this receives UnacceptableExpiryError, do not respond with HTML error 422.
```
lumberj: The wording should change because we're not dealing with HTTP error codes here

lumberj:
LGTM. One minor nit in a comment

PR #122, opened 2016-03-23
Refactor towards three-layer architecture

lumberj:
Create app container
Pass services down into lower layers (lower layers don't require singletons from src/services)

TODO:
- Move common utils (e.g., services/log) to `common`


lumberj:
cc @gip -- config is now passed down from top layer

src/index.js
```diff
@@ -1,65 +1,19 @@
...
-  })
-}
+const connector = createApp(config, backend, ledgers)
 
 module.exports = {
-  app: app,
-  listen: listen,
+  app: connector.koaApp,
+  listen: connector.listen,
   addLedger: ledgers.addLedger.bind(ledgers),
```
MatthewPhinney: Is this doable with with `_.partial`, as in `app.js` (for consistency)?

src/models/payments.js
```diff
@@ -316,13 +313,13 @@ function * settle (payment) {
   }
 }
 
-function isTraderFunds (funds) {
+function isTraderFunds (config, funds) {
```
MatthewPhinney: Any reason why `config` is the first param here. Above it's towards the end of the parameter list.

MatthewPhinney: Any reason why `config` is the first param here? Above it's towards the end of the param list.

src/index.js
lumberj: I just grepped for `addLedger` and it's not used

src/models/payments.js
lumberj: Because later it's used as a callback function, so I was able to do `_.partial(isTraderFunds, config)`

src/index.js
sentientwaffle: It isn't used internally, which is why it is exported.

lumberj: Where is it used? (Curious)

sentientwaffle: https://github.com/ripple/five-bells-connector-lib-ripple

src/models/payments.js
```diff
@@ -334,8 +331,9 @@ function * updateSourceTransfer (updatedTransfer, traderCredit) {
     source_transfers: [updatedTransfer],
     destination_transfers: [destinationTransfer]
   }
-  yield validate(payment)
-  yield settle(payment)
+
+  yield validate(payment, ledgers, config, backend)
```
MatthewPhinney: Do you need to update the signature for `validate` to accept these params?

clark800:
I’m pondering the DI structure. One thing I've noticed is that refactoring things like this can be quicker if you use keyword arguments because for example if you want to delete one of them, you can just delete it where it is used and where it is passed in and you don’t have to change all the intermediate functions that were just passing it through. So we might want to consider using a "context" object to hold all the dependency injection objects. It also gives you a logical separation between the DI parameters and non-DI parameters.

src/models/payments.js
lumberj: https://github.com/interledger/five-bells-connector/pull/122/files#diff-32de0c2324711eaaab279194f31a0160R283

clark800:
I'm not sure if we're getting much advantage to passing the `backend` around as a parameter because the `config` already specifies the backend, so wherever you need the backend you could just call a function that loads the proper backend based on the config. Or am I missing something? Also, I haven't figured out what the `ledgers` parameter is used for yet. I'm wondering if there is a way we can avoid passing that around too.

lumberj:
@clark800 I think maybe I can followup with a later commit to avoid needing to pass some things down. The config specifies a backend type, which is then `require`'ed in services/backend.

emschwartz:
Could you provide more context for the purpose of this refactoring? Is there a specific issue that motivated it or is it just to comply with the 3-layer architecture design?

clark800:
Also, we could even eliminate a lot of the config parameters by passing it at the module level like this

```
const makeExports = (config) => {
  getX: (param) => {
    ... // have access to config here even though it isn't a parameter to the getX function
  }
}

module.exports = makeExports
```

lumberj:
@clark800 I like that approach.
@emschwartz discussion is here https://github.com/interledger/five-bells-ledger/issues/112

MatthewPhinney:
> Is there a specific issue that motivated it or is it just to comply with the 3-layer architecture design?

@emschwartz my understanding was that this was partly motivated by the desire to unit test different quoter backends. With this refactor, we can pass the config as a param into the instantiation of the app, allowing us to switch quoter backends. See [here](https://ripplelabs.atlassian.net/browse/RILP-284). Is this correct @lumberj?

emschwartz:
Sounds like a use case for https://github.com/justmoon/reduct

On Wed, Mar 23, 2016 at 3:15 PM, Matthew Phinney <notifications@github.com>
wrote:

> Is there a specific issue that motivated it or is it just to comply with
> the 3-layer architecture design?
>
> @emschwartz <https://github.com/emschwartz> my understanding was that
> this was partly motivated by the desire to unit test different quoter
> backends. With this refactor, we can pass the config as a param into the
> instantiation of the app, allowing us to switch quoter backends. See here
> <https://ripplelabs.atlassian.net/browse/RILP-284>. Is this correct
> @lumberj <https://github.com/lumberj>?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/interledger/five-bells-connector/pull/122#issuecomment-200565260>
>



-- 
Evan Schwartz | Software Architect | Ripple
[image: ripple.com] <http://ripple.com>

lumberj:
@MatthewPhinney Yes, that's one of a few things

clark800:
Maybe a cleaner way of doing it:

```
module.exports = function (config) {

  function getX (param) {

  }

  function getY (param) {

  }

  return {getX, getY}
}
```

lumberj:
Blocked by https://github.com/interledger/five-bells-notary/pull/44/

clark800:
LGTM

PR #124, opened 2016-03-25
Quote with unsupported ledger should return 422

MatthewPhinney:
- Move common quote methods to `utils`. 
- Return 422 on quote with unsupported currency pair

### DIFF
--- a/schemas/PositiveFiniteAmount.json
+++ b/schemas/PositiveFiniteAmount.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "PositiveFiniteAmount",
+  "title": "PositiveFiniteAmount",
+  "description": "A string representation of a positive, finite number",
+  "type": "string",
+  "pattern": "/^[+]?([1-9][0-9]*(?:[.][0-9]*)?|0*.0*[1-9][0-9]*)(?:[eE][+-][0-9]+)?$/"
+}
--- a/src/backends/fixerio/index.js
+++ b/src/backends/fixerio/index.js
@@ -6,19 +6,9 @@ const BigNumber = require('bignumber.js')
 const AssetsNotTradedError = require('../../errors/assets-not-traded-error')
 const NoAmountSpecifiedError = require('../../errors/no-amount-specified-error')
 const log = require('../../common').log('fixerio')
-const config = require('../../services/config')
+const utils = require('../utils')
 
 const RATES_API = 'https://api.fixer.io/latest'
-
-function lookupCurrencies (source_ledger, destination_ledger) {
-  for (let pair of config.get('tradingPairs')) {
-    if (pair[0].indexOf(source_ledger) === 4 &&
-      pair[1].indexOf(destination_ledger) === 4) {
-      return [pair[0].slice(0, 3), pair[1].slice(0, 3)]
-    }
-  }
-  return null
-}
 
 /**
  * Dummy backend that uses Fixer.io API for FX rates
@@ -67,19 +57,6 @@ class FixerIoBackend {
     this.currencies.sort()
   }
 
-  /**
-   * Check if we trade the given pair of assets
-   *
-   * @param {String} source The URI of the source ledger
-   * @param {String} destination The URI of the destination ledger
-   * @return {boolean}
-   */
-  * hasPair (source, destination) {
-    const currencyPair = lookupCurrencies(source, destination)
-    return _.includes(this.currencies, currencyPair[0]) &&
-      _.includes(this.currencies, currencyPair[1])
-  }
-
   _formatAmount (amount) {
     return new BigNumber(amount).toFixed(2)
   }
@@ -106,15 +83,15 @@ class FixerIoBackend {
    */
   * getQuote (params) {
     // Throw an error if the currency pair is not supported
-    const hasPair = yield this.hasPair(params.source_ledger, params.destination_ledger)
+    const hasPair = utils.hasPair(this.currencies, params.source_ledger, params.destination_ledger)
     if (!hasPair) {
-      console.log('doesnt have pair', params)
+      log.error('doesnt have pair', JSON.stringify(params))
       throw new AssetsNotTradedError('This connector does not support the ' +
         'given asset pair')
     }
 
     // Get ratio between currencies and apply spread
-    const currencyPair = lookupCurrencies(params.source_ledger, params.destination_ledger)
+    const currencyPair = utils.getCurrencyPair(params.source_ledger, params.destination_ledger)
     const destinationRate = this.rates[currencyPair[1]]
     const sourceRate = this.rates[currencyPair[0]]
 
--- a/src/backends/fixerio/index.js
+++ b/src/backends/fixerio/index.js
@@ -3,10 +3,11 @@
 const _ = require('lodash')
 const request = require('co-request')
 const BigNumber = require('bignumber.js')
-const AssetsNotTradedError = require('../../errors/assets-not-traded-error')
 const NoAmountSpecifiedError = require('../../errors/no-amount-specified-error')
 const log = require('../../common').log('fixerio')
+const config = require('../../services/config')
 const utils = require('../utils')
+const ServerError = require('five-bells-shared/errors/server-error')
 
 const RATES_API = 'https://api.fixer.io/latest'
 
@@ -30,6 +31,12 @@ class FixerIoBackend {
 
     this.rates = {}
     this.currencies = []
+    const ledgerPairs = config.get('tradingPairs')
+    if (_.isEmpty(ledgerPairs)) {
+      throw new ServerError('No trading pairs found for this connector')
+    }
+    this.currencyPairs = ledgerPairs.map((p) => [p[0].slice(0, 3),
+                                   p[1].slice(0, 3)])
   }
 
   /**
@@ -50,7 +57,6 @@ class FixerIoBackend {
       })
       apiData = result.body
     }
-
     this.rates = apiData.rates
     this.rates[apiData.base] = 1
     this.currencies = _.keys(this.rates)
@@ -82,14 +88,6 @@ class FixerIoBackend {
    * @param {String|Integer|BigNumber} params.destination_amount The amount of the destination asset we want to send (either this or the source_amount must be set)
    */
   * getQuote (params) {
-    // Throw an error if the currency pair is not supported
-    const hasPair = utils.hasPair(this.currencies, params.source_ledger, params.destination_ledger)
-    if (!hasPair) {
-      log.error('doesnt have pair', JSON.stringify(params))
-      throw new AssetsNotTradedError('This connector does not support the ' +
-        'given asset pair')
-    }
-
     // Get ratio between currencies and apply spread
     const currencyPair = utils.getCurrencyPair(params.source_ledger, params.destination_ledger)
     const destinationRate = this.rates[currencyPair[1]]
@@ -122,8 +120,8 @@ class FixerIoBackend {
     return {
       source_ledger: params.source_ledger,
       destination_ledger: params.destination_ledger,
-      source_amount: sourceAmount,
-      destination_amount: destinationAmount
+      source_amount: sourceAmount.toString(),
+      destination_amount: destinationAmount.toString()
     }
   }
 
--- a/src/backends/ilp-quoter/index.js
+++ b/src/backends/ilp-quoter/index.js
@@ -8,20 +8,10 @@ const NoAmountSpecifiedError =
 const UnacceptableQuoterAmountError =
   require('../../errors/unacceptable-quoter-amount-error')
 const log = require('../../common').log('ilpquote')
+const ServerError = require('five-bells-shared/errors/server-error')
 const config = require('../../services/config')
 const BigNumber = require('bignumber.js')
-
-function lookupCurrencies (source_ledger, destination_ledger) {
-  for (let pair of config.get('tradingPairs')) {
-    if (pair[0].indexOf(source_ledger) === 4 &&
-        pair[1].indexOf(destination_ledger) === 4) {
-      const currencyA = pair[0].slice(0, 3)
-      const currencyB = pair[1].slice(0, 3)
-      return [currencyA, currencyB]
-    }
-  }
-  return null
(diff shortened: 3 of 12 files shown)


src/backends/ilp-quoter/index.js
```diff
@@ -85,6 +75,10 @@ class ILPQuoter {
                 currencyPair[0] + '/' + currencyPair[1] + '/' + amount +
                 '/' + type
     const result = yield request({ uri, json: true })
+    if (result.statusCode >= 400) {
```
MatthewPhinney: @lumberj or @gip do we have a good way to unit test this? I think the tests are using the fixer.io backend, but I could be wrong.

lumberj: Tests are using fixerio backend

src/backends/utils.js
```diff
@@ -0,0 +1,37 @@
+'use strict'
+
+const _ = require('lodash')
+const config = require('../services/config')
+
+function lookupCurrencies (source_ledger, destination_ledger) {
```
lumberj: Maybe `getCurrencyPair` makes more sense?

src/backends/ilp-quoter/index.js
```diff
@@ -85,28 +75,40 @@ class ILPQuoter {
...
     const result = yield request({ uri, json: true })
-    const fixedAmount = type === 'source' ? result.body.source_amount
-                                          : result.body.destination_amount
+    if (result.statusCode >= 400) {
+      log.error('Error getting quote: ', JSON.stringify(result.body))
+      throw new ServerError('Unable to get quote from backend.')
+    }
+    const source_amount = result.body.source_amount
+    const destination_amount = result.body.destination_amount
+    if (!source_amount || !destination_amount) {
```
clark800: Would this be a good case for a JSON schema?

src/backends/ilp-quoter/index.js
```diff
@@ -85,28 +75,40 @@ class ILPQuoter {
...
+    }
+    const fixedAmount = type === 'source' ? source_amount : destination_amount
     if (fixedAmount !== amount.toString()) {
       if (!(new BigNumber(fixedAmount).equals(new BigNumber(amount)))) {
         throw new UnacceptableQuoterAmountError('Backend returned an invalid ' +
                   type + ' amount: ' + fixedAmount + ' (expected: ' + amount + ')')
       }
     }
+
+    if (isNaN(source_amount) || Number(source_amount) === 0 || Number(source_amount) === Number.POSITIVE_INFINITY) {
```
lumberj: Maybe pull some of this validation into a `validate` function. Also, any reason to not use Bignumber here?

src/backends/ilp-quoter/index.js
```diff
@@ -85,28 +75,40 @@ class ILPQuoter {
...
+      log.error('Quoter returned invalid source_amount: ', JSON.stringify(result.body))
+      throw new ServerError('Quoter returned invalid source_amount: ' + result.body.source_amount)
+    }
+    if (isNaN(destination_amount) || Number(destination_amount) === 0 || Number(destination_amount) === Number.POSITIVE_INFINITY) {
```
lumberj: Maybe you want to check for `<=0` so you catch negative numbers too

MatthewPhinney:
> Would this be a good case for a JSON schema?

Yeah, I think a schema would make sense here.

src/backends/ilp-quoter/index.js
lumberj: Should this validation maybe be in models/quote? These validation rules are across any type of backend

src/models/quote.js
```diff
@@ -122,7 +125,26 @@ function * getAccountLedger (account) {
 
 function * getQuote (params, ledgers, config) {
   const query = yield makeQuoteQuery(params, config)
+  const type = query.source_amount ? 'source' : 'destination'
+  const amount = query.source_amount || query.destination_amount
   const quote = yield backend.getQuote(makeQuoteArgs(query))
+  try {
+    validate('Quote', quote)
```
MatthewPhinney: @lumberj added schema validation to the model. I'm re-throwing the error as a server error, since the validator returns an InvalidBodyError. I'm not sure that `ServerError` is the best error here, but I think it should be some kind of 500 error.

schemas/PositiveFiniteAmount.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "PositiveFiniteAmount",
+  "title": "PositiveFiniteAmount",
+  "description": "A string representation of a positive, finite number",
+  "type": "string",
+  "pattern": "^(?!(?:^[-+]?[0.]+(?:[Ee]|$)))(?!(?:^-))(?:(?:[+-]?)(?=[0123456789.])(?:(?:(?:[0123456789]+)(?:(?:[.])(?:[0123456789]*))?|(?:(?:[.])(?:[0123456789]+))))(?:(?:[Ee])(?:(?:[+-]?)(?:[0123456789]+))|))$"
```
lumberj: This looks really complex. What about using https://github.com/ripple/ripple-lib/blob/develop/src/common/schemas/objects/value.json#L7 ?

src/models/quote.js
lumberj: What about adding an `isValid` method to the validator? That way you don't have try/catch/rethrow

test/quoteSpec.js
```diff
@@ -48,8 +48,130 @@ describe('Quotes', function () {
...
+          expect(res.body.id).to.equal('InvalidAmountSpecifiedError')
+          expect(res.body.message).to.equal('destination_amount must be finite and positive')
+        })
+        .end()
+    })
+
+    it('should return a 400 if destination_amount is infinite', function * () {
+      yield this.request()
+        .get('/quote?' +
+          'destination_amount=1e100000&source_ledger=http://eur-ledger.example/EUR' +
```
lumberj: Is this test necessary once we do this? https://github.com/interledger/five-bells-connector/pull/127

src/backends/fixerio/index.js
```diff
@@ -145,8 +122,8 @@ class FixerIoBackend {
     return {
       source_ledger: params.source_ledger,
       destination_ledger: params.destination_ledger,
-      source_amount: sourceAmount,
-      destination_amount: destinationAmount
+      source_amount: sourceAmount.toString(),
+      destination_amount: destinationAmount.toString()
```
MatthewPhinney: @lumberj the reason I changed this to return strings was because it looks like the other backend is returning the amounts as strings, and I wanted consistency across all backends, so I could validate the response with the JSON schema.

test/quoteSpec.js
MatthewPhinney: I rebased off #127 and will clean up the tests.

MatthewPhinney: Here is what happens when quoting with 0/&infin; under #127.

Quote with zero returns:
```
422 UnacceptableAmountError Quoted source is lower than minimum amount
allowed
```
Quote with infinity returns:
```
422 UnacceptableAmountError: Insufficient liquidity in market maker account
```
While this could be considered a valid solution, I think it's best to catch these edge cases and return a 400 error. This avoids polluting the logs with pages of numbers (we log the amount if it exceeds the ledger precision, and logging `1e10000` as a bignumber takes well over a page.

lumberj: > logging 1e10000 as a bignumber takes well over a page).

Objects should be `JSON.stringify()`ed before logging

lumberj: Also, `1e100000` is not Infinity? Why would the "Insufficient liqudity" error be incorrect here?

MatthewPhinney: Yeah, actually that was a hack to test a string that would evaluate to infinity as a js number. In my mind this is just a question of where to put the validation. If we know the quoted amount is invalid, I don't see a point in requesting a quote from the backend for the amount.

lumberj: Agreed, but maybe change the string to `foo` or some other string that doesn't match the schema.

src/backends/utils.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Check if we trade the given pair of assets
+ *
+ * @param {String} source The URI of the source ledger
+ * @param {String} destination The URI of the destination ledger
+ * @return {boolean}
+ */
+function hasPair (currencyPairs, source, destination) {
+  const currencyPair = getCurrencyPair(source, destination)
+  return _.includes(currencyPairs, currencyPair[0]) &&
```
clark800: This looks like it's checking a list of pairs for a component of a pair; that should always be false.

MatthewPhinney: This method was moved from the fixer.io backend: https://github.com/interledger/five-bells-connector/pull/124/files#diff-732a4256af7e30fb991b56df0f8474f2L77. Both the fixerio and ilp-quoter backends had the same method, implemented slightly differently, so I consolidated them in one file.

This method first looks up the currency pair for the `source`/`destination` ledger (something like `['USD', 'EUR']`). `currencyPairs` is an array of the currencies the ledger supports. So we return `true` iff `currencyPairs` contains both the currencies the user wants to trade.

The variable names are probably the confusing part. I'll update them.

src/backends/utils.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Check if we trade the given pair of assets
+ *
+ * @param {Array.<Array.<String>>} Array of pairs
+ *   of currencies traded by this connector
+ * @param {String} source The URI of the source ledger
+ * @param {String} destination The URI of the destination ledger
+ * @return {boolean}
+ */
+function hasPair (currencyPairs, source, destination) {
```
MatthewPhinney: Actually @clark800 nice catch ... it looks like the two backends structured the pairs differently. It was a flat array in fixer.io and an array of arrays (each array being a pair) in ilp-quoter. I changed this to allow the more general case of trading more than one currency pair. All the more reason to have unit tests for the ilp-quoter backend. There is some larger cleanup that could be done with the backends, but that's out of the scope of this PR.

schemas/PositiveFiniteAmount.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "PositiveFiniteAmount",
+  "title": "PositiveFiniteAmount",
+  "description": "A string representation of a positive, finite number",
+  "type": "string",
+  "pattern": "/^[+]?([1-9][0-9]*(?:[.][0-9]*)?|0*.0*[1-9][0-9]*)(?:[eE][+-][0-9]+)?$/"
```
lumberj: https://github.com/interledger/five-bells-shared/blob/master/schemas/Amount.json#L6 is much shorter, and I think you just need to change it to:

```
"^[+]?[1-9]*[.]?[0-9]+([eE][-+]?[0-9]+)?$"
```

src/backends/fixerio/index.js
```diff
@@ -60,26 +51,15 @@ class FixerIoBackend {
       })
       apiData = result.body
     }
-
+    const pairs = config.get('tradingPairs')
+    this.pairs = pairs.map((p) => [p[0].slice(0, 3),
```
lumberj: It's confusing to assign `this.pairs` to a thing that is different than `pairs`. Maybe this should be called `currencyCodePairs`?

src/backends/fixerio/index.js
```diff
@@ -60,26 +51,15 @@ class FixerIoBackend {
       })
       apiData = result.body
     }
-
+    const pairs = config.get('tradingPairs')
```
clark800: Is this guaranteed to be set? Even if so it might be good to check here to make sure.

src/backends/fixerio/index.js
lumberj:
```
this.currencyPairs = ledgerPairs.map(...)
```

src/backends/ilp-quoter/index.js
```diff
@@ -61,13 +48,13 @@ class ILPQuoter {
...
     log.debug('Connecting to ' + this.backendUri)
-    const hasPair = yield this.hasPair(params.source_ledger,
+    const hasPair = utils.hasPair(this.pairs, params.source_ledger,
                                        params.destination_ledger)
     if (!hasPair) {
       throw new AssetsNotTradedError('This connector does not support the ' +
         'given asset pair')
     }
-    const currencyPair = lookupCurrencies(params.source_ledger,
+    const currencyPair = utils.getCurrencyPair(params.source_ledger,
```
lumberj: You can probably cleanup utils.hasPair because it's calling utils.getCurrencyPair, and everywhere you use one you're using the other one.

So maybe just have a `utils.hasCurrencyPair` and use `utils.getCurrencyPair`

Is utils.hasPair necessary? If utils.getCurrencyPairs returns an empty pair `[]` then you know that `hasPair` is false, right?

src/controllers/quote.js
```diff
@@ -109,6 +112,25 @@ const model = require('../models/quote')
 /* eslint-enable */
 
 exports.get = function * () {
+  const queryParams = this.query
```
lumberj: Maybe have a `function validateAmounts (query.source_amount, query.destination_amount)`, just to make the `get` function a bit smaller

schemas/PositiveFiniteAmount.json
MatthewPhinney: That will validate against `0` and not against `0.1`. I'm all for simplifying the regex, but I haven't found anything shorter that does what I need. On the other hand, we could allow `0` here, since the amount is validated here: https://github.com/interledger/five-bells-connector/blob/master/src/models/quote.js#L158. That may be easiest.

lumberj: Let's keep it the way you have it. It will fail sooner

src/backends/fixerio/index.js
lumberj: It's guaranteed to be something, but I don't think it's validated in the config

MatthewPhinney: Yeah, it's possible this is empty, which is why I had to add a check here: https://github.com/interledger/five-bells-connector/pull/124/files#diff-9439739e49d60f21e447a7bdcc0ef546R7. I can do the same here.

clark800:
LGTM

PR #125, opened 2016-03-25
BREAKING | FEATURE: Connector verifies notification signatures

lumberj:
Ledger signing public keys for notifications.
Public keys are stored on disk in PEM format

Breaking:
- In production Connector requires notification signing keys


lumberj:
cc @emschwartz @justmoon

src/lib/config.js
```diff
@@ -101,16 +137,37 @@ function validateLocalEnvConfig () {
...
+  if (notifications.must_verify) {
+    const ledgers = parseLedgers()
+    for (let ledgerObj of ledgers) {
+      const uri = ledgerObj.ledger
+      if (notifications.keys[uri] === undefined) {
+        throw new Error(`Missing notification signing keys for ledger: ${uri}`)
+      }
+
+      try {
+        fs.accessSync(notifications.keys[uri], fs.R_OK)
```
MatthewPhinney: So the point of this call is to prevent us from having to a `try`/`catch` block when reading the file?

lumberj: Yes. The validation is separate from parsing the config object

test/configSpec.js
```diff
@@ -277,5 +285,74 @@ describe('ConnectorConfig', function () {
...
+        })
+      })
+
+      describe('CONNECTOR_NOTIFICATIONS_VERIFY=false', () => {
+        beforeEach(() => {
+          process.env.CONNECTOR_NOTIFICATIONS_VERIFY = '0'
+        })
+
+        it('config.notifications.must_verify=false', () => {
+          const config = loadConnectorConfig()
```
lumberj: This test should have NODE_ENV=production, otherwise it defaults to false in non-prod

test/configSpec.js
```diff
@@ -277,5 +285,74 @@ describe('ConnectorConfig', function () {
...
+        const config = loadConnectorConfig()
+        expect(config.get('notifications.must_verify')).to.equal(false)
+      })
+
+      describe('CONNECTOR_NOTIFICATION_VERIFY=true', () => {
+        beforeEach(() => {
+          process.env.CONNECTOR_NOTIFICATION_VERIFY = '1'
+        })
+
+        it('config.notifications.must_verify=false', () => {
```
MatthewPhinney: Don't you want `must_verify=true`?

src/controllers/notifications.js
```diff
@@ -71,6 +73,16 @@ const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
 exports.post = function * postNotification () {
   const notification = yield requestUtil.validateBody(this, 'Notification')
   log.debug('Got notification: ' + JSON.stringify(notification))
+
+  const verifySignature = this.config.get('notifications.must_verify')
+  if (verifySignature) {
+    const result = model.verifySignature(notification, this.config)
+    if (!result.valid) {
+      log.warn('Signature verification failed: ' + result.error)
+      throw new UnprocessableEntityError('Notification failed signature verification')
```
naoitoi: Would this cause the scalability problem we have addressed recently?  
1. Ledger -> Connector : notification with wrong signature
2. Ledger <- Connector : 422 because the signature is wrong
3. Repeat from step 1. 
Flooding the servers and impacting the performance negatively?

lumberj: > Flooding the servers and impacting the performance negatively?

In a production environment, you'd probably block a server that is trying to the connector with bad requests

src/lib/config.js
```diff
@@ -63,7 +63,41 @@ function parseAdminEnv () {
...
+  // e.g. ["USD@http://usd-ledger.example","EUR@http://eur-ledger.example/some/path"]
+  return JSON.parse(Config.getEnv(envPrefix, 'LEDGERS') || '[]').map((ledger) => {
+    const sep = ledger.indexOf('@')
+    return { currency: ledger.substr(0, sep), ledger: ledger.substr(sep + 1) }
+  })
+}
+
+function parseNotificationSignEnv () {
+  // On by default in prod
+  const isProduction = process.env.NODE_ENV === 'production'
```
clark800: I think we should try to have just one line of code in the whole application that looks at `process.env`, then we could easily override this without mutating `process.env`

src/lib/config.js
```diff
@@ -101,16 +137,37 @@ function validateLocalEnvConfig () {
...
+function validateNotificationEnv () {
+  // Validate notification signing public keys
+  const notifications = parseNotificationSignEnv()
+
+  if (notifications.must_verify) {
+    const ledgers = parseLedgers()
+    for (let ledgerObj of ledgers) {
+      const uri = ledgerObj.ledger
+      if (notifications.keys[uri] === undefined) {
+        throw new Error(`Missing notification signing keys for ledger: ${uri}`)
```
naoitoi: I see some throw new Error().  Do all these Errors result in Connector not starting?  
Could you tell us the thoughts behind using throw new Error() as supposed to returning an error or throwing a more specific error?

src/lib/config.js
```diff
@@ -181,12 +240,13 @@ function getLocalConfig () {
...
     server,
-    backendUri
+    backendUri,
+    notifications
   }
 }
 
 function loadConnectorConfig () {
-  return Config.loadConfig(envPrefix, getLocalConfig())
+  return Config.loadConfig(envPrefix, getLocalConfig(), {ed25519: false})
```
naoitoi: What does setting "ed25519" to false do?

README.md
```diff
@@ -87,10 +87,20 @@ npm start
   ]
 ]
 ```
+
+* `CONNECTOR_NOTIFICATION_VERIFY` (default: `'true'`) The connector verifies the signatures on notifications.
```
naoitoi: If Connector's running in production, I wonder if this should be configurable.
It seems to me that, in production, this should be always enforced - Connector must always verify notifications.

src/lib/config.js
lumberj: Sure. The `loadConfig` exported function could take an object that overrides that. That can be a followup refactor PR

src/lib/config.js
lumberj: The connector will fail to start, so it's not necessary to create a typed error here, just a good error message.

src/lib/config.js
lumberj: The config module in five-bells-shared currently tries to parse ed25519 keys for all components. In production it fails if they don't exist. This is a workaround to avoid releasing a breaking change in five-bells-shared right now. I have a followup to revisit this in the next sprint (RILP-409)

README.md
lumberj: That's why we default to `true`. Is there a reason to make this non-configurable?

src/lib/config.js
lumberj: https://ripplelabs.atlassian.net/browse/RILP-411

src/models/metadata.js
```diff
@@ -2,7 +2,6 @@
 
 function getMetadata (config) {
   return {
-    public_key: config.getIn(['keys', 'ed25519', 'public']),
```
clark800: Why was this removed?

lumberj: The Connector does not use this

lumberj: It's not configured with ed25519 keys, which are used by the ledger and notary for signing crypto conditions. I think this must have been a copy/paste error

src/lib/config.js
clark800: OK

test/configSpec.js
```diff
@@ -277,5 +285,75 @@ describe('ConnectorConfig', function () {
...
+          const config = loadConnectorConfig()
+          expect(config.get('notifications.must_verify')).to.equal(true)
+        })
+
+        it('throws if missing public key for all ledgers', () => {
+          process.env.CONNECTOR_NOTIFICATION_KEYS = undefined
```
emschwartz: What if `CONNECTOR_NOTIFICATION_VERIFY` is false?

test/data/notificationSigned.json
```diff
@@ -0,0 +1,36 @@
...
+    "debits":[{
+      "amount":"1.00",
+      "account":"http://eur-ledger.example/accounts/mark"
+    }],
+    "credits":[{
+      "amount":"1.00",
+      "account":"http://eur-ledger.example/accounts/bob"
+    }],
+    "state": "proposed"
+  }]
```
emschwartz: This looks like a payment that is not signed, maybe it should be called something other than `notificationSigned`

test/notificationSpec.js
```diff
@@ -56,7 +87,7 @@ describe('Notifications', function () {
       delete this.notificationNoConditionFulfillment.id
       yield this.request()
         .post('/notifications')
-        .send(this.notificationNoConditionFulfillment)
+        .send(signNotification(this.notificationNoConditionFulfillment))
```
emschwartz: I don't think you should sign every notification in the tests. It seems like it adds an unnecessary step to every single test. What about instead just having one set of tests where it requires the notifications to be signed and then tests just the functionality you want?

README.md
naoitoi: If somebody accidentally sets CONNECTOR_NOTIFICATION_VERIFY to false in production (e.g. somebody sets it to false in proof of concept, and forgets to reset it when it moves to production), it would allow an attacker to increase his or her balance on the ledger.

src/lib/config.js
naoitoi: OK.

test/configSpec.js
lumberj: I can add that test case

test/data/notificationSigned.json
lumberj: Dang, this file shouldn't be in here. From an older revision

src/controllers/notifications.js
naoitoi: I can see both arguments:
On one hand, the situation where Ledger's sending an invalid signature is so bad (certificate expired, private key misconfigured, etc.) that maybe it's fine to fail dramatically. 
On the other hand, unfortunately, these things tend to happen in real life.  When it happens, we wouldn't want to flood the servers. 
How about logging an error and returning 200, like we do here?
https://github.com/interledger/five-bells-connector/blob/master/src/controllers/notifications.js#L81

lumberj: > When it happens, we wouldn't want to flood the servers. 
> How about logging an error and returning 200, like we do here?

There's no way to know anything is failing unless there's a separate process parsing the logs. I prefer explicitly returning an error code and letting the caller know that something is wrong.

README.md
lumberj: I don't buy this argument. They need to explicitly set this to false. They could also make a number of other bad decisions, like not running HTTPS, or incorrectly configuring the public keys.

test/notificationSpec.js
lumberj: ok

test/configSpec.js
lumberj: Done

test/notificationSpec.js
lumberj: Done

src/controllers/notifications.js
MatthewPhinney: > There's no way to know anything is failing unless there's a separate process parsing the logs.

This is also true in our current implementation of the ledger. Ledger gets error, logs, retries notification, gets error, logs, retries, etc. As far as I know, we don't have a mechanism for informing RC of a failed notification between the ledger and the connector. The errors bounce back and forth between the two servers and are confined to the logs.

The most obvious way this issue could arise is if the connector/ledger are configured with different keys. This is an exceptional case, and should probably result in an exception, but we should handle this, rather than blindly retrying the notification, as we currently do.

My understanding is that notification retries address the situation in which one server is down temporarily or, for whatever reason, the notification didn't reach the recipient, which is not the case here. In this case, it seems unlikely that retrying the notification will result in a different outcome, since notifications are signed the same way each time.

One way to address this would be to use a more specific error than `UnprocessableEntityError` (maybe `InvalidSignatureError` or something), so that we can treat instances of this error differently. I think the above code works for the PR, but we shouldn't lose sight of this discussion.

MatthewPhinney:
LGTM

test/notificationSpec.js
```diff
@@ -844,92 +913,92 @@ describe('Notifications', function () {
...
-        .reply(200, {})
-
-      yield this.request()
-        .post('/notifications')
-        .send(this.notificationSourceTransferAtomic)
-        .expect(200)
-        .end()
-    })
-  })
+    describe('atomic mode: one case', function () {
```
emschwartz: It looks like the test changes below here should be in a different PR (or the PR description should be updated to reflect these changes)

lumberj: I didn't change anything, it's just the diff of things moving.

src/controllers/notifications.js
lumberj: > My understanding is that notification retries address the situation in which one server is down temporarily or, for whatever reason, the notification didn't reach the recipient, which is not the case here. 

Right, so if the ledger gets a 500 error, something was wrong with the server, it should retry. If it gets a 4XX error, something is wrong with the client (ledger) and it should fix the problem before retrying.

naoitoi: A little bit of context:
@justmoon earlier made a point that Ledger should retry on all errors (4xx and 5xx) because there might be a component, e.g., a firewall, between Ledger and Connector.  That component might respond to Ledger with an error for a reason Connector doesn't control.  If that happens, Ledger should keep retrying (with backoff), to be on the cautious side.  So that's why Ledger retries on both 4xx and 5xx today.  
But - I see @lumberj 's point, too.  In this case, Ledger failed to deliver a good notification with a good signature.  Ledger did something wrong, and it should receive an error.  
So I agree with @MatthewPhinney - this code is fine for now, but let's remember we'll have a message flooding problem if this error condition gets triggered (could you leave a comment in code?)  If this becomes a real problem, in the future, we can do what @MatthewPhinney suggested - throw a different error and treat it differently.

naoitoi:
LGTM.  This was a big change with a lot of back and forth - Great job Alan!

src/controllers/notifications.js
lumberj: > If that happens, Ledger should keep retrying (with backoff), to be on the cautious side. So that's why Ledger retries on both 4xx and 5xx today.

The ledger retrying on 4XX errors seems to be a bug according to the definitions of 4XX status codes.

> 10.4.1 400 Bad Request

> The 400 (Bad Request) status code indicates that the server cannot or
   will not process the request due to something that is perceived to be
   a client error (e.g., malformed request syntax, invalid request
   message framing, or deceptive request routing).

Source: https://tools.ietf.org/html/rfc7231#section-6.5

> 11.2.  422 Unprocessable Entity

   > The 422 (Unprocessable Entity) status code means the server
   understands the content type of the request entity (hence a
   415(Unsupported Media Type) status code is inappropriate), and the
   syntax of the request entity is correct (thus a 400 (Bad Request)
   status code is inappropriate) but was unable to process the contained
   instructions.  For example, this error condition may occur if an XML
   request body contains well-formed (i.e., syntactically correct), but
   semantically erroneous, XML instructions.

I haven't seen a proxy server or load balancer return a 4XX when the upstream server is at fault, that's usually what 502, 503, and 504 are reserved for. For example, if the Connector tries to `PUT transfer` on the upstream ledger and fails (gets a 5XX), it should return a `502`. I'm not sure if it does that today.

justmoon: 422 seems like the correct code to return. (401 would be more specific, but we're not using a standard HTTP authentication method, so it doesn't work in our case.)

Regarding retry behavior: Generally, we retry on 4xx even though it goes against HTTP. We do this because we would like to be conservative. We should have reasonable backoff and intelligent scheduling to avoid hammering a server to death. But the protocol very heavily relies on reliable delivery, so giving up on a notification should be a very last resort.

> I haven't seen a proxy server or load balancer return a 4XX when the upstream server is at fault, that's usually what 502, 503, and 504 are reserved for.

It commonly happens when a load balancer is pointing at the wrong server, so the server is responding with 404 or 403 because it doesn't know the URL. It could also be a misconfigured component, for instance the connector will respond 422 if it doesn't recognize the asset pair, which would happen if it is temporarily loading the wrong config.

We could stop retrying if we see InvalidKeyError, but I think that's the exact opposite of what we should do. The scenario I'm imagining is that a ledger somehow has a misconfigured key. Their notifications start failing, so they get notified. They go in and fix the configuration and restart the server. If we make the behavior non-retry, they now have to somehow reset the state of the failed notifications. If we make the behavior retry, they don't have to do anything, things will just work themselves out.

justmoon: > If the load balancer or upstream server is misconfigured, retrying won't help.

Sure retrying will help. The next time you retry, the misconfiguration may have been fixed.

> Seeing many failed payments, though, might be an indicator that something went wrong.

With atomic mode, infinite retries and proper monitoring (each 4xx error between banks for instance should cause an immediate ops alert), there should never **be** a single prepared payment that fails.

> You could also add metadata to each failed notification indicating if it was retried and only retry on startup (i.e.g, ledger is running, notifications fail, in the db with retry=false. Ledger restarts, loads failed notifications with retry=false, re-sends them, sets retry=true).

That only helps if the misconfiguration is on the ledger side, not if its on the connector side.

lumberj: From, https://github.com/interledger/five-bells-connector/pull/125#discussion_r57749447

>  but let's remember we'll have a message flooding problem if this error condition gets triggered 

cc @naoitoi 
So, in reality, we should never have a message flooding problem.

Retrying infinite times is probably the right thing to do then. If we assume that owners of Ledgers, and other components are rational, knowledgeable operators. If a Connector is flooded with messages by a Ledger they trust, these are problems that the software can't fix.

PR #127, opened 2016-03-28
BREAKING | FEATURE: Get precision from ledger

lumberj:
Requires the ledger to support returning its precision
https://github.com/interledger/five-bells-ledger/commit/87f8622e25ae2b0d3ea9c39d7284c00aaa5b4e83

TODO:
- Caching ledger precision

### DIFF
--- a/src/models/quote.js
+++ b/src/models/quote.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const request = require('co-request')
+const _ = require('lodash')
 const BigNumber = require('bignumber.js')
 const log = require('../common').log('quote')
 const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
@@ -9,6 +10,31 @@ const InvalidURIParameterError = require('five-bells-shared').InvalidUriParamete
 const ExternalError = require('../errors/external-error')
 const balanceCache = require('../services/balance-cache.js')
 const backend = require('../services/backend')
+
+function * getPrecisionAndScale (ledger) {
+  log.debug('getPrecisionAndScale', ledger)
+  function throwErr () {
+    throw new ExternalError('Unable to determine ledger precision')
+  }
+
+  let res
+  try {
+    res = yield request(ledger, {json: true})
+  } catch (e) {
+    if (!res || res.statusCode !== 200) {
+      log.debug('getPrecisionAndScale', e)
+      throwErr()
+    }
+  }
+
+  if (!res || res.statusCode !== 200) throwErr()
+
+  log.debug('getPrecisionAndScale', res.body)
+  return {
+    precision: res.body.precision,
+    scale: res.body.scale
+  }
+}
 
 function * makeQuoteQuery (params, config) {
   // TODO: include the expiry duration in the quote logic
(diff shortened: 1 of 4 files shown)


lumberj:
cc @emschwartz

lumberj:
cc @MatthewPhinney

src/lib/utils.js
```diff
@@ -1,19 +1,21 @@
...
-   *
-   * @param {array} arr Input array
-   * @return {array[]} Possible pairs
-   */
-  static getPairs (arr) {
-    return arr.reduce((prev, cur, i) => (
-      prev.concat(arr.slice(i + 1).map((val) => [cur, val]))
-    ), [])
-  }
+/**
```
MatthewPhinney: This is also changed in #125: https://github.com/interledger/five-bells-connector/pull/125/files#diff-fb54f78b07de422efaa3d2d82bb82db3L3.

emschwartz:
LGTM

src/models/quote.js
```diff
@@ -10,6 +11,32 @@ const ExternalError = require('../errors/external-error')
...
+// TODO: Validate amounts are not greater than ledger precision
+function * getPrecisionAndScale (ledger) {
+  log.debug('getPrecisionAndScale', ledger)
+  function throwErr () {
+    throw new ExternalError('Unable to determine ledger precision')
+  }
+
+  let res
+  try {
+    res = yield request(ledger, {json: true})
```
MatthewPhinney: Under what circumstances will `request` throw? I thought it would just return an error code if something goes wrong.

src/models/quote.js
```diff
@@ -66,7 +93,7 @@ function * makeQuoteQuery (params, config) {
   }
 }
 
-function makeQuoteArgs (query) {
+function * makeQuoteArgs (query) {
```
MatthewPhinney: Why does this now return a generator?

src/models/quote.js
lumberj: On lower level socket errors I think.

src/lib/utils.js
lumberj: Git will sort it out

src/models/quote.js
```diff
@@ -10,6 +11,32 @@ const ExternalError = require('../errors/external-error')
 const balanceCache = require('../services/balance-cache.js')
 const backend = require('../services/backend')
 
+// TODO: Validate amounts are not greater than ledger precision
```
lumberj: Oh I can remove this

src/models/quote.js
lumberj: Not sure how that `*` snuck in there

src/models/quote.js
```diff
@@ -120,23 +146,50 @@ function * getAccountLedger (account) {
...
+  const roundedDestinationAmount = new BigNumber(quote.destination_amount).toFixed(
+    dstPrecisionAndScale.scale, BigNumber.ROUND_DOWN)
+
+  validatePrecision(roundedSourceAmount, sourcePrecisionAndScale.precision, 'source')
+  validatePrecision(roundedDestinationAmount, dstPrecisionAndScale.precision, 'destination')
+
+  const roundedQuote = _.assign(_.clone(quote), {
+    source_amount: roundedSourceAmount,
+    destination_amount: roundedDestinationAmount
+  })
```
emschwartz: What do you think the behavior should be if either the source or destination amounts are rounded down to zero? Should it return a quote with an amount set to 0? Should it ensure that both numbers are at least the minimum allowed by the ledger?

lumberj: > Should it ensure that both numbers are at least the minimum allowed by the ledger?

That makes sense. How can the Connector know what the minimum amount is? Should it check that it's at least a positive amount, maybe?

MatthewPhinney: I think checking for a positive amount would be good.

lumberj: Done

lumberj: The source_amount cannot ever round down to zero, since it's using the `ROUND_UP` rule

test/quoteSpec.js
```diff
@@ -126,6 +126,19 @@ describe('Quotes', function () {
...
 
+    it('should return a 422 if destination_ledger rounded amount is less than or equal to 0', function * () {
+      yield this.request()
+        .get('/quote?source_amount=0.00001' +
+          '&source_ledger=http://eur-ledger.example/EUR' +
+          '&destination_ledger=http://usd-ledger.example/USD')
+        .expect(422)
+        .expect(function (res) {
+          expect(res.body.id).to.equal('UnacceptableAmountError')
+          expect(res.body.message).to.equal('Quoted destination amount exceeds minimum amount allowed')
```
emschwartz: "Exceeds" means "is greater than"

lumberj: Or "to go beyond a permitted limit". Do you have a suggestion for a
different word to use?

lumberj: I'll just say "Quoted destination amount is less than minimum amount allowed"

lumberj: Done

PR #129, opened 2016-03-29
[FEATURE] Cache ledger precision

lumberj:


package.json
```diff
@@ -43,6 +43,7 @@
     "koa-route": "^2.4.2",
     "koa-static": "^1.4.5",
     "lodash": "^4.6.1",
+    "lru-cache": "^4.0.1",
```
naoitoi: Since this data never expires, how about we simply store it in a variable?  
Saving introducing a new package.

src/models/quote.js
```diff
@@ -36,6 +37,22 @@ function * getPrecisionAndScale (ledger) {
   }
 }
 
+const LRU = require('lru-cache')
+const cache = LRU({
+  max: 100,
+  maxAge: 1 / 0, // Infinity. Never expire
```
MatthewPhinney: Does `Number.POSITIVE_INFINITY` work here? I think that's a bit more clear.

lumberj: Sure

package.json
lumberj: ok

lumberj:
Going to close, fix tests, and address comments

lumberj:
Replaced with #132

PR #130, opened 2016-03-29
Fix 500 error on quote request for disallowed pair

clark800:
Tested by Nguyen


lumberj:
LGTM

MatthewPhinney:
I actually addressed this in #124 because I didn't realize we had a separate issue for it. I cleaned it up as part of a larger refactor.

PR #131, opened 2016-03-30
[Docker] Connector runs on 3000 by default

lumberj:


emschwartz:
What's the logic for changing it to 3000?

lumberj:
> What's the logic for changing it to 3000?

The shared config module defaults to 3000
https://github.com/interledger/five-bells-shared/blob/master/lib/config.js#L83

If the operator wants to change the exposed port they can use 

```
CONNECTOR_PORT=5000 \
  docker run interledger/five-bells-connector \
  --expose $CONNECTOR_PORT \
  --env $CONNECTOR_PORT
```

MatthewPhinney:
LGTM

clark800:
Maybe we should have different default ports for the different components?

lumberj:
@clark800  makes sense. Can followup in the future

PR #132, opened 2016-03-30
[FEATURE] Cache ledger precision

lumberj:


test/quoteSpec.js
```diff
@@ -362,6 +364,44 @@ describe('Quotes', function () {
         .end()
     })
 
+    it('caches source and destination ledger precision', function * () {
+      nock.cleanAll()
+      nock('http://eur-ledger.example/EUR')
+        .get('').reply(200, {precision: 10, scale: 4})
+        .get('').reply(500, 'Invalid request')
```
clark800: What's this for?

clark800: Can you use nock's `done` method instead?

clark800:
LGTM

test/quoteSpec.js
lumberj: I don't think that does what we want here. I don't see any way in nock to create a negative expectation.

clark800: Won't nock already fail if you make more requests than are listed?

lumberj: Yea, that's the point here. It shouldn't make the second request.

clark800: I mean I don't think you need to list the 500

clark800: https://github.com/node-nock/nock/issues/415

lumberj: Ah right, yea it might not be necessary, but this makes it explicit

lumberj: I can't find docs for `done` behavior

PR #134, opened 2016-03-30
[FIX] Include port in ledger URI

MatthewPhinney:


MatthewPhinney:
@lumberj what do you think of including the port info in the URI for the ledger? We need this for the ILP server tests, since all ledgers run on local host. If we don't have this, we can't specify different keys for different ledgers (localhost:3000 and localhost:4001 would be considered the same).

lumberj:
@MatthewPhinney it's allowed in the URI spec, so should be fine. https://tools.ietf.org/html/rfc3986#section-3.2.3

clark800:
LGTM

test/notificationSpec.js
```diff
@@ -1013,6 +1014,13 @@ describe('Notifications', function () {
       })
     })
   })
+
+  describe('parseLedger', function () {
```
lumberj: What about adding a testing with a ledger that has a port number, instead of unit testing a private function?

MatthewPhinney: Oops, already merged. I can add that test case, but I wanted to release this and get ILP server working, so that I can finish adding the signing keys to the server by the end of the day.

PR #137, opened 2016-04-01
[FEATURE] Enable CORS

emschwartz:
Ledger [allows CORS](https://github.com/interledger/five-bells-ledger/blob/master/src/lib/app.js#L38) but the connector does not

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "five-bells-shared": "^13.0.0",
     "koa": "^1.0.0",
     "koa-compress": "^1.0.6",
+    "koa-cors": "0.0.16",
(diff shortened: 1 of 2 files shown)


lumberj:
Why is this necessary?

emschwartz:
If the sender is used within the browser it needs to be able to communicate with the connectors to ask for quotes

lumberj:
LGTM

clark800:
> If the sender is used within the browser it needs to be able to communicate with the connectors to ask for quotes

This knowledge will probably get lost in the PR comments; might be a good thing to put in a code comment.

PR #138, opened 2016-04-01
[FEATURE] move pathfinder to connector

sentientwaffle:
Corresponding sender changes: https://github.com/interledger/five-bells-sender/pull/56

This isn't breaking because its behind a feature flag by default.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -84,6 +84,7 @@ function createApp (config, ledgers) {
   koaApp.use(route.get('/pairs', pairs.getCollection))
 
   koaApp.use(route.get('/quote', quote.get))
+  koaApp.use(route.get('/quote_local', quote.getLocal))
 
   koaApp.use(route.post('/notifications', notifications.post))
 
--- a/src/models/quote.js
+++ b/src/models/quote.js
@@ -4,6 +4,7 @@ const request = require('co-request')
 const _ = require('lodash')
 const BigNumber = require('bignumber.js')
 const log = require('../common').log('quote')
+const Pathfinder = require('five-bells-pathfind').Pathfinder
 const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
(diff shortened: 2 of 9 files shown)


lumberj:
Update the routes here, https://github.com/interledger/five-bells-connector/blob/master/src/models/metadata.js#L9

src/app.js
```diff
@@ -84,6 +84,7 @@ function createApp (config, ledgers) {
   koaApp.use(route.get('/pairs', pairs.getCollection))
 
   koaApp.use(route.get('/quote', quote.get))
+  koaApp.use(route.get('/quote_local', quote.getLocal))
```
lumberj: Why not use a query parameter `?local=true` instead?

src/models/quote.js
```diff
@@ -197,6 +198,42 @@ function * getQuote (params, ledgers, config) {
...
+ * @param {String} params.source_account
+ * @param {String} params.source_amount
+ * @param {String} params.destination_ledger
+ * @param {String} params.destination_account
+ * @param {String} params.destination_amount
+ * @param {Object} config
+ * @returns {Quote[]}
+ */
+function * getQuotePath (params, config) {
+  // TODO cache pathfinder so that it doesn't have to re-crawl for every payment
```
emschwartz: Can't wait for this change :) -- that'll improve the performance by a lot

emschwartz:
LGTM

sentientwaffle:
@lumberj: I'll add it to metadata.

Re:query param: `/quote_local` has a slightly different return type. And it doesn't really matter one way or the other -- it's only a temporary endpoint. It will no longer be needed once distance-vector routing is implemented.

lumberj:
>  It will no longer be needed once distance-vector routing is implemented.

What is that? Maybe you can add some context into the commit message.

PR #140, opened 2016-04-01
fix: allow using a CA with basic-auth credentials

lumberj:
Previously, CA file was only parsed correctly when using client cert
authentication.

This correctly parses the file, which is useful if using self-signed
certificates in testing environments


clark800:
LGTM

sentientwaffle:
To get the integration tests to pass rebase onto master.

PR #141, opened 2016-04-01
Allows empty list of currency pairs

gip:


gip:
cc @lumberj and @MatthewPhinney

src/backends/fixerio/index.js
```diff
@@ -32,9 +31,6 @@ class FixerIoBackend {
     this.rates = {}
     this.currencies = []
     const ledgerPairs = config.get('tradingPairs')
-    if (_.isEmpty(ledgerPairs)) {
```
MatthewPhinney: This was a "safety check" that Chris and I decided to add. Live and learn, I guess ...

MatthewPhinney:
LGTM

src/backends/fixerio/index.js
clark800: I meant that we should check for `null`/`undefined`, sorry I should have caught this in the review

clark800: Still think we should check for `null`/`undefined` though

gip: Makes sense. I'll add that when I work on the unit tests.

PR #142, opened 2016-04-05
Unit test for ilpquoter class

gip:


src/backends/utils.js
```diff
@@ -12,22 +10,25 @@ const log = require('../common').log('fixerio')
...
-  if (!_.isEmpty(config.get('tradingPairs'))) {
-    for (let pair of config.get('tradingPairs')) {
-      // trading pair is of the form
-      // ["<currency@<source_ledger>","<currency>@<destination_ledger>"]
-      if (pair[0].indexOf(sourceLedger) === 4 &&
-        pair[1].indexOf(destinationLedger) === 4) {
-        return [pair[0].slice(0, 3), pair[1].slice(0, 3)]
+function getCurrencyPairFromConfig (config) {
+  const log = require('../common').log(config.get('backend'))
+  return function getCurrencyPair (sourceLedger, destinationLedger) {
```
clark800: Could you just use partial function application instead of a function that returns a function?

gip: What's the problem with returning a closure?

clark800: Partial application returns a closure too, just in a simpler way

test/ilpQuoterSpec.js
```diff
@@ -0,0 +1,110 @@
...
+      const scope = nock(this.BACKEND_URI)
+                      .put('/pair/EUR/USD').reply(200)
+                      .put('/pair/USD/EUR').reply(200)
+      yield this.backend.connect()
+      scope.isDone()
+    })
+
+    it('should fail for a quote with a missing source_ledger', function * () {
+      const quote = { source_amount: '123', destination_ledger: 'https://localhost:4000' }
+      yield yieldAndAssertException(this.backend.getQuote(quote), AssetsNotTradedError)
```
MatthewPhinney: The `yieldAndAssertException` method looks fine to me, but I noticed, in other places in the codebase (example: [here](https://github.com/interledger/five-bells-shared/blob/707263764d36ae0da8b3ec0150845ce8703293a6/test/configSpec.js#L85)), when we want to check for a specific error, we generally use
```
expect(() => { this.backend.getQuote(quote) }).to.throw(AssetsNotTradedError)
```
For consistency, I think it makes sense to use the same style here.

src/backends/fixerio/index.js
```diff
@@ -85,7 +85,7 @@ class FixerIoBackend {
    */
   * getQuote (params) {
     // Get ratio between currencies and apply spread
-    const currencyPair = utils.getCurrencyPair(params.source_ledger, params.destination_ledger)
+    const currencyPair = this.getCurrencyPair(params.source_ledger, params.destination_ledger)
```
lumberj: Why does this function need to be a method on the object? Can you just call `utils.getCurrencyPairFromConfig` here?

src/backends/ilp-quoter/index.js
```diff
@@ -16,11 +15,12 @@ const utils = require('../utils')
  *   precision. So amounts are passed using JSON strings
  */
 class ILPQuoter {
-  constructor (opts) {
+  constructor (config, opts) {
     log.debug('ILPQuoter ctor')
     const ledgerPairs = config.get('tradingPairs')
+    this.getCurrencyPair = utils.getCurrencyPairFromConfig(config)
```
lumberj: Same comment as above

src/backends/ilp-quoter/index.js
```diff
@@ -29,10 +29,14 @@ class ILPQuoter {
    *   and checks that all the pairs are supported
    */
   * connect () {
-    const requests = this.currencyPairs.map((pair) => {
+    const uris = []
+    this.currencyPairs.forEach((pair) => {
```
lumberj: Can this be a separate function `getCurrenyPairUris()`?

```js
function getCurrencyPairURIs (currencyPairs, backendUri) {
  return _.uniq(_.compact(_.map(currencyPairs, (pair) = {
    return backendUri + '/pair/' + pair[0] + '/' + pair[1];
  })))
}
```

src/backends/ilp-quoter/index.js
```diff
@@ -29,10 +29,14 @@ class ILPQuoter {
...
       const uri = this.backendUri + '/pair/' + pair[0] + '/' + pair[1]
-      return request({ method: 'PUT', uri, json: true })
+      if (uris.indexOf(uri) === -1) {
+        uris.push(uri)
+      }
     })
+    const requests = uris.map((uri) => request({ method: 'PUT', uri, json: true }))
```
lumberj: What happens if one of these responses is not 200?

src/services/backend.js
```diff
@@ -8,6 +8,6 @@ if (!Backend) {
     'module specified by CONNECTOR_BACKEND was not found in /backends')
 }
 
-module.exports = new Backend({
+module.exports = new Backend(config, {
```
lumberj: Instead of passing the entire config, can you pass in just the required config values, similar to how fxSpread is being passed in?

test/ilpQuoterSpec.js
lumberj: I think the issue is that co-mocha doesn't let you use that `expect` with generator functions.

I mentioned using `co` because mocha works just fine with promises.

```
co.wrap(this.backend.getQuote)(quote).catch((err) => expect(err)....
```

src/backends/fixerio/index.js
gip: I could call `getCurrencyPair` and pass the `config`, `source_ledger` and `destination_ledger`
Is it what you are suggesting?

src/backends/utils.js
```diff
@@ -12,22 +10,25 @@ const log = require('../common').log('fixerio')
...
-function getCurrencyPair (sourceLedger, destinationLedger) {
-  if (!_.isEmpty(config.get('tradingPairs'))) {
-    for (let pair of config.get('tradingPairs')) {
-      // trading pair is of the form
-      // ["<currency@<source_ledger>","<currency>@<destination_ledger>"]
-      if (pair[0].indexOf(sourceLedger) === 4 &&
-        pair[1].indexOf(destinationLedger) === 4) {
-        return [pair[0].slice(0, 3), pair[1].slice(0, 3)]
+function getCurrencyPairFromConfig (config) {
+  const log = require('../common').log(config.get('backend'))
```
lumberj: `require` statements should be at the top of the file

src/backends/fixerio/index.js
lumberj: Yes. And it avoids a function that returns a function.

src/backends/ilp-quoter/index.js
lumberj: I meant, is there a unit test for that case

src/services/backend.js
lumberj: I think the only other config value that's requires is the backend uri

lumberj: This makes unit tests cleaner too, because it doesn't require mutating the environment and loading the config

src/backends/utils.js
gip: Can you write something like what is below using `bind()`?
```function f (a,b) {
  const x = // some expression of a and b
  return function (c) { 
    return // some expression of c and x
  }
} ```

Comment (clark800) [inline src/backends/utils.js]:
```
function f (a, b, c) {
  const x = // some expression of a and b
  return // some expression of c and x
}
const f1 = _.partial(f, a1, b1)
const result = f1(c1)
```

Comment (lumberj) [inline src/backends/utils.js]: Is this really a case where all this is necessary? All the arguments to the inner function are known when the outer function is called

Comment (gip) [inline src/backends/utils.js]: Was there a decision that high-order functions / partial application should not be used in our code? From a functional programming viewpoint it's a totally valid way of doing this kind of things.

Comment (lumberj) [inline src/backends/utils.js]: Sure, use higher order functions. The name of the function should indicate it's returning another function. 

One change you can make though is to add the function as a property of the prototype, not the instance.

File: src/backends/ilp-quoter/index.js (hunk 6 above)
Comment (lumberj) [inline src/backends/ilp-quoter/index.js]: ?

File: src/services/backend.js (hunk 7 above)
Comment (gip) [inline src/services/backend.js]: We want a ctor with a common interface though. The quoter needs to know `backendUri` while fixerio does not. It seems better to keep `config` to abstract the different parameters.

Comment (lumberj) [inline src/services/backend.js]: The fixerio backend will just ignore options it doesn't care about

File: src/backends/utils.js (hunk 1 above)
Comment (gip) [inline src/backends/utils.js]: Finally I decided to change the code not to use HOF since it doesn't seem to be an agreement on what approach to use HOF vs partial app vs function.

File: test/ilpQuoterSpec.js (hunk 2 above)
Comment (gip) [inline test/ilpQuoterSpec.js]: I couldn't make the code work with co.wrap() : the test wasn't running. Let's fix that in a next PR.

File: src/backends/fixerio/index.js (hunk 3 above)
Comment (gip) [inline src/backends/fixerio/index.js]: Done - passing the necessary argument without HOF

File: src/backends/ilp-quoter/index.js (hunk 6 above)
Comment (gip) [inline src/backends/ilp-quoter/index.js]: This has been fixed

File: src/services/backend.js (hunk 7 above)
Comment (gip) [inline src/services/backend.js]: Fixed

File: test/ilpQuoterSpec.js [hunk 9]
```diff
@@ -0,0 +1,129 @@
+'use strict'
+
+const nock = require('nock')
+nock.enableNetConnect(['localhost'])
```
clark800: What's this for? Are you making actual socket connections?

src/backends/utils.js
gip: Fixed

src/backends/fixerio/index.js
```diff
@@ -30,8 +29,8 @@ class FixerIoBackend {
 
     this.rates = {}
     this.currencies = []
-    const ledgerPairs = config.get('tradingPairs')
-    this.currencyPairs = ledgerPairs.map((p) => [p[0].slice(0, 3),
+    this.pairs = opts.pairs
```
clark800: Since there is also "currencyPairs", is there a more specific name for this?

lumberj:
LGTM

src/backends/fixerio/index.js
gip: Basically `pairs` is on the form `[["USD@https://host.com/", "EUR@https://host2.com"]]` and `currencyPairs` will be `[["USD", "EUR"]]` so it's different variables

clark800: Right, the comment was about the variable _naming_ though

test/ilpQuoterSpec.js
gip: I actually do not know. I saw that at other locations in the code and just included that change as I thought it must have some usefulness. See for instance: https://github.com/interledger/five-bells-notary/blob/master/test/healthSpec.js#L4

src/backends/fixerio/index.js
gip: Right. Would `ledgerPairs` work? Or `currencyWithLedgerPairs` ?

test/ilpQuoterSpec.js
clark800: Ok, please be careful not to write code that you don't understand. This is to allow nock to make actual socket connections. If you are not using it then I think it should be removed.

src/backends/fixerio/index.js
clark800: This is a tough one... `currencyWithLedgerPairs` seems reasonable

test/ilpQuoterSpec.js
gip: Ok, though this is not functional code, it's tests written on top of a DSL and I don't think I want to spend too much time understanding that stuff in details. It's second order with respect to our goal and an imprecise and abstracted understanding of the capabilities is totally ok with me. I'll remove that since *you* know it's not necessary.

gip:
@clark800 Previous comment addressed

clark800:
LGTM

PR #146, opened 2016-04-23
New crypto-conditions

justmoon:
This migrates Five Bells Connector to the new crypto-conditions format. See https://github.com/interledger/five-bells-condition

Requires interledger/five-bells-ledger#233, interledger/five-bells-integration-test#17


src/models/payments.js
```diff
@@ -352,21 +309,20 @@ function * updateDestinationTransfer (updatedTransfer, traderDebit, relatedResou
 }
 
 function * updateTransfer (updatedTransfer, relatedResources, ledgers, config) {
-  // Maybe it's a source transfer:
+  // Maybe it's a destination transfer:
+  const traderDebit = updatedTransfer.debits.find(_.partial(isTraderFunds, config))
+  if (traderDebit) {
+    yield updateDestinationTransfer(updatedTransfer, traderDebit, relatedResources)
```
MatthewPhinney: No `return`?

src/lib/testPaymentExpiry.js
```diff
@@ -37,12 +37,6 @@ TransferTester.prototype.isAtomic = function () {
   return getTransferCases(this.source_transfers[0]).length > 0
 }
 
-TransferTester.prototype.isFinal = function () {
-  return !_.some(this.destination_transfers, function (transfer) {
-    return transfer.hasOwnProperty('execution_condition')
-  })
-}
-
```
emschwartz: https://github.com/interledger/five-bells-connector/blob/feature/st-new-cc/src/lib/testPaymentExpiry.js#L113-L147 is also about using the last ledger's transfer state receipt as the condition

emschwartz:
LGTM

src/controllers/notifications.js
```diff
@@ -86,16 +90,47 @@ exports.post = function * postNotification () {
...
+    if (
+      (e instanceof AssetsNotTradedError) ||
+      (e instanceof NoRelatedDestinationDebitError) ||
+      (e instanceof NoRelatedSourceCreditError) ||
+      (e instanceof UnacceptableConditionsError) ||
+      (e instanceof UnacceptableExpiryError) ||
+      (e instanceof UnacceptableRateError) ||
+      (e instanceof UnrelatedNotificationError)
+    ) {
+      // Certain exceptions indicate an error in the transfer memo, rather than
```
gip: // Certain exceptions indicate an error in the transfer memo, rather than
      // the transfer itself. Since the client is the ledger and the ledger is
      // not at fault here, it doesn't make sense to return a 4xx response. In
      // other words, both the ledger and connector have done everything right,
      // so the request status between them should be 200.

Status code 202 (The request has been accepted for processing, but the processing has not been completed) could also makes sense here -- it is a 'success' from the requester's perspective but also indicates some issue with the request.

src/controllers/notifications.js
```diff
@@ -86,16 +90,47 @@ exports.post = function * postNotification () {
   try {
     yield model.processNotification(notification, this.ledgers, this.config)
   } catch (e) {
-    if (!(e instanceof UnacceptableExpiryError)) {
+    if (
+      (e instanceof AssetsNotTradedError) ||
```
clark800: Could these be put under a parent class so there is just one class to check for?

clark800:
LGTM

src/controllers/notifications.js
justmoon: Great idea, I'll add that.

src/controllers/notifications.js
justmoon: Good idea, but would defer to a separate PR in the interest of progress.

src/controllers/notifications.js
justmoon: After reading about it "202 Accepted" seems to be more for requests that haven't been processed yet, which isn't quite the case here. If we wanted to return 202, we should probably return it always and actually do the processing after the request. I'll leave the decision about switching to 202 for a future PR.

PR #147, opened 2016-04-27
Use crypto-conditions for notification signing

justmoon:
Using crypto-conditions speeds up notification signing and makes it more flexible (multi-sig, algorithm selection, etc.)

Depends on interledger/five-bells-ledger#235


emschwartz:
LGTM (once tests pass)

gip:
LGTM

PR #150, opened 2016-05-02
[BREAKING] destination routing

sentientwaffle:
Requires https://github.com/interledger/five-bells-shared/pull/134
Requires https://github.com/interledger/five-bells-routing/pull/2

Corresponding sender changes: https://github.com/interledger/five-bells-sender/pull/65


lumberj:
@sentientwaffle or @justmoon is there a document for this feature?

src/controllers/routes.js
```diff
@@ -0,0 +1,23 @@
...
+
+const requestUtil = require('five-bells-shared/utils/request')
+const routingTables = require('../services/routing-tables')
+const routeBroadcaster = require('../services/route-broadcaster')
+const knownConnectors = {}
+
+exports.post = function * () {
+  const routes = yield requestUtil.validateBody(this, 'Routes')
+
+  // TODO verify that POSTer of these routes matches route.connector.
```
emschwartz: @sentientwaffle @justmoon How are connectors going to verify routes are coming from the right connector?

src/controllers/routes.js
```diff
@@ -0,0 +1,23 @@
...
+  for (const route of routes) {
+    routingTables.addRoute(route)
+  }
+
+  const connector = routes[0] && routes[0].connector
+  if (connector && !knownConnectors[connector]) {
+    yield routeBroadcaster.broadcast()
```
emschwartz: This will only come back from the yield after all the notifications have been sent out to other connectors, shouldn't it respond to the connector that sent it this route update before sending out the other notifications?

src/lib/route-broadcaster.js
```diff
@@ -0,0 +1,126 @@
...
+      }
+    })
+  }
+
+  crawlLedgers () {
+    return Object.keys(this.adjacentLedgers).map(this._crawlLedger, this)
+  }
+
+  * _crawlLedger (ledger) {
+    const res = yield request({
```
emschwartz: This should use the multiledger abstraction (if you don't want to change it here we can also make that change when everything is switched over to using the ledger plugin framework)

src/lib/route-builder.js
```diff
@@ -0,0 +1,162 @@
...
+   * @param {String} query.destinationLedger
+   * @param {String} query.destinationAccount
+   * @param {String} query.destinationAmount
+   * @returns {Transfer}
+   */
+  * getQuote (query) {
+    const _nextHop = this._findNextHop(query)
+    if (!_nextHop) throwAssetsNotTradedError()
+    if (query.sourceAmount) {
+      _nextHop.finalAmount = (+_nextHop.finalAmount * (1 - this.slippage)).toFixed(15)
```
emschwartz: Why is 15 hard-coded here?

src/lib/route-builder.js
```diff
@@ -0,0 +1,162 @@
...
+      }],
+      _hop: nextHop
+    }
+  }
+
+  /**
+   * @param {Transfer} sourceTransfer
+   * @returns {Transfer} destinationTransfer
+   */
+  * getDestinationTransfer (sourceTransfer) {
```
emschwartz: Might be helpful to add jsdoc descriptions for these functions to explain what they do. This function signature isn't very intuitive

src/models/payments.js
```diff
@@ -251,13 +245,21 @@ function isTraderFunds (config, funds) {
...
-  const isTransferReady = updatedTransfer.state === 'prepared' || updatedTransfer.state === 'executed'
+  const isTransferReady = updatedTransfer.state === 'prepared' ||
+    (updatedTransfer.state === 'executed' && !updatedTransfer.execution_condition)
   if (!isTransferReady) return
 
+  if (!config.ledgerCredentials[updatedTransfer.ledger]) {
+    throw new AssetsNotTradedError('This connector does not support the given asset pair')
+  }
+
+  // TODO this is cheating, but how do we know the type of the final transfer's ledger?
```
clark800: Why is this cheating? Why do you need to know the type of the final transfer's ledger? What is a "type" of a ledger and how does this relate to the code below?

emschwartz: `five-bells-ledger` is a type of ledger. Bitcoin is another type of ledger. We want to the connector to be able to work with any type of ledger so this means that all ledger interactions should go through the [`multiledger` abstraction](https://github.com/interledger/five-bells-connector/blob/master/src/lib/ledgers/multiledger.js).

I think @sentientwaffle said that is cheating because we can't actually assume that a transfer on a different ledger will have the `ledger` field. I think we will need to have some kind of wrapper around transfers from other ledgers to make them easier to work with (i.e. `{ ledger: 'bitcoin', tx: '...' }` and maybe pull out some of the other fields as a convenience)

clark800:
Not fully understanding everything, but I didn't notice any real issues. I'm not sure if it is necessary to use classes here, it may be simpler to use functions since the classes don't seem to own any state, but not a blocker. LGTM after tests pass.

src/controllers/routes.js
sentientwaffle: Well, eventually routes will arrive on a transfer's `additional_info` (as opposed to the dedicated `/routes` endpoint), so the sending connector's identity would be implicitly verified when they authorize the debit.

src/controllers/routes.js
sentientwaffle: You're right.

src/lib/route-broadcaster.js
sentientwaffle: Ya, its so far behind I'll just wait until we switch everything over.

src/lib/route-builder.js
sentientwaffle: 15 was because according to [bignumber.js](http://mikemcl.github.io/bignumber.js/#bignumber):
> Values of type number with more than 15 significant digits are considered invalid 

I'll change it to just use bignumber for the math, though.

src/models/payments.js
sentientwaffle: Yep, the connector needs to know the `ledger` of the final transfer so that it knows how to route it.

src/controllers/routes.js
emschwartz: Hmm, we'll have to talk more about that. I'm not sure it makes sense to add routes to the incoming transfer. Even if we want to have connectors pay one another it might make more sense to pre-pay a bit and draw down against a balance rather than pay for every advertisement.

src/lib/config.js
```diff
@@ -195,12 +194,14 @@ function getLocalConfig () {
...
   expiry.minMessageWindow =
-    Config.getEnv(envPrefix, 'MIN_MESSAGE_WINDOW') || 1 // seconds
-  expiry.maxHoldTime = Config.getEnv(envPrefix, 'MAX_HOLD_TIME') || 10 // seconds
+    +Config.getEnv(envPrefix, 'MIN_MESSAGE_WINDOW') || 1 // seconds
+  expiry.maxHoldTime = +Config.getEnv(envPrefix, 'MAX_HOLD_TIME') || 10 // seconds
 
   // The spread is added to every quoted rate
   const fxSpread = Number(Config.getEnv(envPrefix, 'FX_SPREAD')) || 0.002 // = .2%
 
+  const slippage = +Config.getEnv(envPrefix, 'SLIPPAGE') || 0.001 // = 0.1%
```
emschwartz: Could you put these hard coded defaults up at the top of the file as `const DEFAULT_MIN_MESSAGE_WINDOW` or something like that?

src/lib/route-broadcaster.js
```diff
@@ -0,0 +1,126 @@
...
+      const destinationLedger = pair[1].split('@')[1]
+      this.adjacentLedgers[destinationLedger] = true
+    }
+  }
+
+  * start () {
+    yield this.crawlLedgers()
+    yield this.reloadLocalRoutes()
+    yield this.broadcast()
+    setInterval(() => this.routingTables.removeExpiredRoutes(), 1000)
```
emschwartz: Seems like this interval should be configurable and/or have a default value defined at the top of the file

src/lib/route-broadcaster.js
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  _tradingPairToQuote (pair) {
+    const sourceLedger = pair[0].split('@')[1]
+    const destinationLedger = pair[1].split('@')[1]
+    // TODO change the backend API to return curves, not points
+    return co(this.backend.getQuote.bind(this.backend), {
+      source_ledger: sourceLedger,
+      destination_ledger: destinationLedger,
+      source_amount: 100000000
```
emschwartz: Why `100000000`? What if the backend says that's too much and doesn't return anything. I think I can see why you might be doing this here but then we should definitely make the `change the backend API to return curves, not points` soon

sentientwaffle: It creates a curve `[[0, 0], [100000000, fx(100000000)]]`. It is just a workaround until the backends are rewritten to return the curve directly.

src/lib/route-builder.js
```diff
@@ -0,0 +1,165 @@
...
+    const sourceExpiryTime = (new Date(sourceExpiry)).getTime()
+    const minMessageWindow = this.minMessageWindow * 1000
+    return (new Date(sourceExpiryTime - minMessageWindow)).toISOString()
+  }
+
+  * _roundHop (hop) {
+    hop.sourceAmount = yield this._roundAmount('source', hop.sourceLedger, hop.sourceAmount)
+    hop.destinationAmount = yield this._roundAmount('destination', hop.destinationLedger, hop.destinationAmount)
+    hop.finalAmount = yield this._roundAmount('destination', hop.finalLedger, hop.finalAmount)
+    return hop
```
emschwartz: What is this function doing?

sentientwaffle: It rounds the hop's amounts according to the corresponding ledgers' scales/precisions. I'll add a comment to that effect in the code as well.

src/lib/route-builder.js
```diff
@@ -0,0 +1,165 @@
...
+          }
+        }
+      }],
+      _hop: nextHop
+    }
+  }
+
+  /**
+   * Given a source transfer with an embedded final transfer, get the next
+   * transfer in the chain.
```
emschwartz: Not strictly necessary for this PR but it would be good to have some docs that explains in detail how this works

src/lib/route-builder.js
emschwartz: Won't some of those amounts be `undefined` when this is called by `getQuote`? Doesn't BigNumber throw if you pass it `null` or `undefined`?

sentientwaffle: That function is called with the result of `_findNextHop`, not the query. All of the amounts are strings at this point.

src/lib/routing-tables.js
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  /**
+   * @param {Route} route
+   */
+  addRoute (route) {
+    const ledgerB = route.source_ledger
+    const ledgerC = route.destination_ledger
+    const connectorFromBToC = route.connector
+    const routeFromBToC = new routing.Route(route.points)
```
emschwartz: Please add more of a description of what this function is doing. What are ledgers A, B, and C? Might be helpful to use more descriptive names

src/lib/routing-tables.js
```diff
@@ -0,0 +1,219 @@
...
+
+    this.eachSource((tableFromA, ledgerA) => {
+      // Don't create A→B→A.
+      if (ledgerA === ledgerC) return
+      const routesFromAToB = tableFromA.destinations.get(ledgerB)
+      if (!routesFromAToB) return
+      // Don't create local route A→B→C if local route A→C already exists.
+      if (this.baseURI === connectorFromBToC && this._getLocalRoute(ledgerA, ledgerC)) return
+      const routeFromAToB = routesFromAToB.get(this.baseURI)
+      // Don't create A→C→B when there is no local A→C.
```
emschwartz: Is the `routeFromBToC` necessarily a two-way route?

emschwartz: Should this say `A->B->C`? I don't quite understand how this code has anything to do with creating a route `A→C→B`

src/lib/routing-tables.js
```diff
@@ -0,0 +1,219 @@
...
+    const nextHop = this._findBestHopForDestinationAmount(ledgerA, ledgerC, +finalAmount)
+    if (!nextHop) return
+    const ledgerB = nextHop.info.nextLedger
+    const routeFromAToB = this._getLocalRoute(ledgerA, ledgerB)
+    const isLocal = this.baseURI === nextHop.bestHop
+    return {
+      connector: nextHop.bestHop,
+      sourceLedger: ledgerA,
+      // Prevent 'BigNumber Error: new BigNumber() number type has more than 15 significant digits'
+      sourceAmount: nextHop.bestCost.toFixed(15),
```
emschwartz: If you call the BigNumber constructor with a string you're not limited in the amount of precision (from [docs](https://mikemcl.github.io/bignumber.js/#bignumber): "Values of type number with more than 15 significant digits are considered invalid (if ERRORS is true) as calling toString or valueOf on such numbers may not result in the intended value. **There is no limit to the number of digits of a value of type string** (other than that of JavaScript's maximum array size).")

src/lib/routing-tables.js
```diff
@@ -0,0 +1,219 @@
...
+      finalAmount: nextHop.bestValue.toFixed(15),
+      minMessageWindow: nextHop.info.minMessageWindow
+    }
+  }
+}
+
+;[
+  'findBestHopForSourceAmount',
+  'findBestHopForDestinationAmount'
+].forEach(function (fn) {
```
emschwartz: This seems like a janky way of calling these two functions. Why not just separate this out into two calls?

src/lib/routing-tables.js
```diff
@@ -0,0 +1,219 @@
...
+  RoutingTables.prototype['_' + fn] = function (source, destination, amount) {
```
emschwartz: Why are you modifying the `RoutingTables.prototype` here?

sentientwaffle: Its my (apparently obfuscated) way of concisely defining two similar methods simultaneously. I'll just dup the code and add them normally.

src/models/quote.js
```diff
@@ -29,211 +56,77 @@ function * makeQuoteQuery (params, config) {
...
  * @param {String} params.source_account
  * @param {String} params.source_amount
+ * @param {String} params.source_expiry_duration
  * @param {String} params.destination_ledger
  * @param {String} params.destination_account
  * @param {String} params.destination_amount
+ * @param {String} params.destination_expiry_duration
  * @param {Object} config
- * @returns {Quote[]}
+ * @returns {Transfer}
```
emschwartz: If I request a quote with a fixed source amount, how can I tell what the destination amount is?

sentientwaffle: `quote.credits[0].memo.destination_transfer` will have the destination amount.

test/quoteSpec.js
```diff
@@ -344,21 +344,6 @@ describe('Quotes', function () {
         .end()
     })
 
-    it('should return a valid Payment Template object', function * () {
```
emschwartz: Can you replace this test with one that validates it against whatever schema it should match?

src/controllers/quote.js
```diff
@@ -113,19 +113,13 @@ const InvalidAmountSpecifiedError = require('../errors/invalid-amount-specified-
...
-  this.body = yield model.getFullQuote(this.query, this.ledgers, this.config)
-}
-
-exports.getLocal = function * () {
-  validateAmounts(this.query.source_amount, this.query.destination_amount)
-  this.body = yield model.getLocalQuote(this.query, this.ledgers, this.config)
+  if (!this.query.destination_account && !this.query.destination_ledger) {
+    throw new InvalidUriParameterError('Missing required parameter: destination_ledger or destination_account')
+  }
+  this.body = yield model.getFullQuote(this.query, this.config)
```
MatthewPhinney: Since `src/models/quote.js` only exposes one method, it may be clearer to call this `getQuote`.

test/quoteSpec.js
```diff
@@ -838,110 +825,72 @@ describe('Quotes', function () {
...
-            expiry_duration: '11'
-          }],
-          destination_transfers: [{
-            ledger: 'http://random-ledger.example/',
-            debits: [{ amount: '200', account: 'http://random-ledger.example/accounts/mary' }],
-            credits: [{ amount: '200', account: null }],
-            expiry_duration: '10'
-          }]
-        })
+        .expect(200)
```
emschwartz: Could you include what the response should look like?

test/routeBroadcasterSpec.js
```diff
@@ -0,0 +1,113 @@
...
+          destination_ledger: 'http://ledgerA.example',
+          connector: 'http://connector.example',
+          min_message_window: 1,
+          source_account: 'http://ledgerB.example/accounts/mark',
+          points: [ [0, 0], [100, 200] ]
+        }
+      ]).reply(200)
+
+      yield this.broadcaster.crawlLedgers()
+      yield this.broadcaster.broadcast()
```
emschwartz: You should add `nock.done()` at the end

test/routeBuilderSpec.js
```diff
@@ -0,0 +1,286 @@
...
+                ledger: ledgerC,
+                debits: [{ account: null, amount: '24.75' }],
+                credits: [{ account: carlC, amount: '24.75' }]
+              }
+            }
+          }],
+          _hop: quoteTransfer._hop
+        })
+      })
+    })
```
emschwartz: If I get multiple quotes from multiple connectors back for a fixed source amount quote, how do I compare them? Do I need to look at `credit[0].memo.destination_transfer.credits[0].amount`? That seems pretty inconvenient...

emschwartz:
Ok, finally made it through the whole thing. The routing changes look good to me overall (there are some details I think could be changed/improved).

The main issue I have with this PR is the change to the connector quote API. I think we might want to have a dedicated quote object that includes details like source and final (recipient) amounts, the expiry time or window, and the transfer (in whatever format the source ledger expects). Especially when we start dealing with different transfer object types it seems like it would be useful to return a consistent object that gives you some details, along with the (opaque) transfer object that you should submit to your ledger to kick the process off. (The only problem I could see with this is you might want to verify that the transfer isn't asking for more money than you wanted to send, or something like that)

I may be missing something but:
* does the sender have to dig into the nested transfers to add the conditions or do connectors infer that the condition for the transfer they create on their right should be the same as the incoming one?
* how does the sender enforce the expiry time and window for the final transfer?

In general, we need to think through how to make sure the recipient of a payment can validate everything they need to about it to avoid having connectors modifying the payment instructions and getting away with it. In theory it doesn't matter if a connector modifies the payment in transit because the sender can check the transfer they're sending to the first connector and the recipient should check the incoming transfer details before signing the receipt. In order to make sure no one gets screwed with this though, we need to be very clear about what the recipient must check.

cc @sentientwaffle @justmoon

src/lib/routing-tables.js
sentientwaffle: You're right, I'll change those to `toString()`.

test/routeBroadcasterSpec.js
sentientwaffle: I have `afterEach(function * () { assert(nock.isDone()) })`, so it covers all of the tests.

test/routeBuilderSpec.js
sentientwaffle: Yes, see [`getCheaperQuote`](https://github.com/interledger/five-bells-sender/pull/65/files#diff-795d9d410a3d31edd97b5216a81441baR41).

Regardless of what non-5B transfer object the connector returns, the sender needs to understand how to compare them.

Maybe it could end up as part of the multiledger abstraction.

src/controllers/routes.js
justmoon: Paying for advertisements should be a standard "paid HTTP" call. So let's make sure we record these requirements (e.g. "how do we know the advertisement is coming from a certain connector") for when we work on the paid-HTTP protocol.

It seems like what would happen is that the other connector would pay you (associating a sending identity with a token with a balance) and then use the token to pay for API calls. If that's the case, then you would be able to associate the token back to the sending identity.

justmoon:
I agree with @emschwartz. Right now the connector returns the transfer object for convenience, but this is probably not a good idea.

First of all, the transfer object is going to look different for each type of ledger. It's an object from the ledger layer, whereas the quoting happens on the Interledger layer. So we're mixing concerns here. Secondly, the sender should carefully check every field in the transfer object. By returning a pre-filled object, we risk that the sender might not check anything or - if they do - they still might forget to check for any unexpected *additional* fields.

What we should return then in the quote request is the minimum required information using ILP standard formats. In addition to that, there should be a field for ledger-specific extensibility.

cc @sentientwaffle

justmoon:
After discussing with @sentientwaffle we decided that since #150 is a blocker for other things and changing the quote response would take until tomorrow, we should merge this now and do the quote response change in a separate PR.

LGTM after tests pass.

PR #151, opened 2016-05-04
[FIX] Do not include basic auth header if pass/user is undefined

gip:


lumberj:
LGTM after tests pass

src/lib/ledgers/five-bells-ledger.js
```diff
@@ -99,15 +99,17 @@ FiveBellsLedger.prototype.subscribe = function * (listener) {
 
   const streamUri = accountUri.replace('http', 'ws') + '/transfers'
   log.debug('subscribing to ' + streamUri)
-  const auth = this.credentials.username + ':' + this.credentials.password
-  const ws = new WebSocket(streamUri, {
-    headers: {
+  const auth = this.credentials.password && this.credentials.username &&
```
clark800: I think the ternary operator is more clear than chaining `&&`

PR #152, opened 2016-05-12
[BREAKING] Dont use Transfers as Quotes

sentientwaffle:
Use the object found in `schemas/Quote.json` for Quotes instead of `Transfer`.

This will break any code that calls `/quote` directly.

The corresponding sender change is https://github.com/interledger/five-bells-sender/pull/69

### DIFF
--- a/schemas/Quote.json
+++ b/schemas/Quote.json
@@ -4,6 +4,10 @@
   "description": "A quote object",
   "type": "object",
   "properties": {
+    "source_connector_account": {
+      "description": "the connector's account on the source ledger",
+      "$ref": "Iri.json"
+    },
     "source_ledger": {
       "description": "The URI of the source ledger",
       "$ref": "Iri.json"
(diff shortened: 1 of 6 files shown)


schemas/Quote.json
```diff
@@ -4,6 +4,10 @@
   "description": "A quote object",
   "type": "object",
   "properties": {
+    "source_connector_account": {
+      "description": "the connector's account on the source ledger",
+      "$ref": "Iri.json"
+    },
     "source_ledger": {
```
emschwartz: Is it necessary to return the source and destination ledgers (or the amount that you sent in the request)?

sentientwaffle: No, I suppose not. I'll remove the ledgers. I think leaving both amounts in is simpler, though.

emschwartz:
Note that this PR also removes the same-ledger quoting functionality from the connector

sentientwaffle:
Ya, that logic is moving to the sender.

emschwartz:
LGTM

justmoon:
LGTM

@sentientwaffle is on vacay - so merging this for him.

PR #154, opened 2016-05-16
Check ledgers' health before crawling / broadcasting routes

gip:
* Check health of all ledgers before to start broadcasting routes
* /health endpoint now returns some actual status on the backend and the routing

### DIFF
--- a/src/backends/ilp-quoter/index.js
+++ b/src/backends/ilp-quoter/index.js
@@ -9,6 +9,7 @@ const UnsupportedPairError =
 const log = require('../../common').log('ilpquoter')
 const ServerError = require('five-bells-shared/errors/server-error')
 const utils = require('../utils')
+const healthStatus = require('../../common/health.js')
 
 /**
  * Example backend that connects to an external component to get
@@ -24,6 +25,7 @@ class ILPQuoter {
     this.currencyPairs = this.currencyWithLedgerPairs.map((p) => [p[0].slice(0, 3),
                                                                   p[1].slice(0, 3)])
     this.backendUri = opts.backendUri
+    this.backendStatus = healthStatus.statusNotOk
   }
 
   * putPair (uri) {
@@ -41,7 +43,18 @@ class ILPQuoter {
   * connect () {
     const uris = _.uniq(this.currencyPairs.map((pair) => this.backendUri + '/pair/' + pair[0] + '/' + pair[1]))
     yield uris.map((uri) => this.putPair(uri))
+    this.backendStatus = healthStatus.statusOk
     return
+  }
+
+  /**
+   * Get backend status
+   */
+  * getStatus () {
+    const status = {
+      backendStatus: this.backendStatus
+    }
+    return status
   }
 
   /**
--- a/src/controllers/health.js
+++ b/src/controllers/health.js
@@ -1,5 +1,8 @@
 'use strict'
 
+const _ = require('lodash')
+const healthStatus = require('../common/health.js')
+
 /**
  * @api {get} /health Get server health status
  * @apiName GetHealth
@@ -14,6 +17,12 @@
  * @returns {void}
  */
 exports.getResource = function * health () {
-  // TODO: Add some checks, e.g. database status
-  this.body = {'status': 'OK'}
+  const backendStatus = yield this.backend.getStatus()
+  const ledgersStatus = yield this.ledgers.getStatus()
+  const body = _.extend({}, backendStatus, ledgersStatus)
+  body.status = (backendStatus.backendStatus === healthStatus.statusOk &&
+                 ledgersStatus.ledgersStatus === healthStatus.statusOk) ? healthStatus.statusOk
+                                                                        : healthStatus.statusNotOk
+  // TODO: add more status, ledger connection, quoter connection,...
+  this.body = body
 }
--- a/src/lib/ledgers/multiledger.js
+++ b/src/lib/ledgers/multiledger.js
@@ -1,12 +1,14 @@
 'use strict'
 
 const FiveBellsLedger = require('./five-bells-ledger')
+const healthStatus = require('../../common/health.js')
 
 function Multiledger (options) {
   this.config = options.config
   this.log = options.log
   this.ledger_types = {undefined: FiveBellsLedger}
-  this.ledgers = {}
+  this.ledgers = this.buildLedgers()
+  this.ledgersHealth = { ledgersHealth: healthStatus.statusNotOk }
 }
 
 Multiledger.prototype.addLedger = function (Ledger) {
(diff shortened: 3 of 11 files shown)


src/backends/ilp-quoter/index.js
```diff
@@ -24,6 +24,7 @@ class ILPQuoter {
     this.currencyPairs = this.currencyWithLedgerPairs.map((p) => [p[0].slice(0, 3),
                                                                   p[1].slice(0, 3)])
     this.backendUri = opts.backendUri
+    this.backendStatus = 'KO'
```
MatthewPhinney: `OK`

gip: Well, this is in the ctor. Initially status is `KO` and becomes `OK` when the quoter has successful responded. Maybe `KO` isn't the best choice -- what would you use?

clark800: I remember seeing "KO" somewhere recently, but I agree it is confusing. How about "NOT READY"?

gip: Sure. Or how about "waiting" so that we don't have a space?

src/controllers/health.js
```diff
@@ -14,6 +16,11 @@
  * @returns {void}
  */
 exports.getResource = function * health () {
-  // TODO: Add some checks, e.g. database status
-  this.body = {'status': 'OK'}
+  const backendStatus = yield this.backend.getStatus()
+  const ledgersStatus = yield this.ledgers.getStatus()
+  const status = _.extend({}, backendStatus, ledgersStatus)
```
clark800: Call this `body`?

src/backends/ilp-quoter/index.js
MatthewPhinney: I think `WAITING` works. I was confused by `KO`.

gip: I already changed it to `NOT READY`. I can always change that later.

MatthewPhinney: It would make sense to be consistent with whatever is chosen here also: https://github.com/interledger/five-bells-connector/pull/154/files#diff-fb64b0ab1de372089a451029c4bc9b18R6.

src/lib/ledgers/multiledger.js
```diff
@@ -7,32 +7,46 @@ function Multiledger (options) {
   this.log = options.log
   this.ledger_types = {undefined: FiveBellsLedger}
   this.ledgers = {}
+  this.buildLedgers()
+  this.ledgersHealth = { ledgersHealth: 'KO' }
```
clark800: Did you mean to change this too?

src/lib/route-broadcaster.js
```diff
@@ -33,12 +35,16 @@ class RouteBroadcaster {
...
   * start () {
     yield this.crawlLedgers()
-    yield this.reloadLocalRoutes()
-    yield this.broadcast()
-    setInterval(() => this.routingTables.removeExpiredRoutes(), CLEANUP_INTERVAL)
+    try {
+      yield this.reloadLocalRoutes()
+      yield this.broadcast()
+    } catch (e) {
+      // This will be retried later
```
clark800: It would be safer to check for a specific exception type and re-throw otherwise

src/lib/ledgers/multiledger.js
```diff
@@ -7,32 +7,46 @@ function Multiledger (options) {
...
 }
 
 Multiledger.prototype.getType = function (ledgerId) {
   return this.getLedger(ledgerId).constructor.TYPE
 }
 
+Multiledger.prototype.checkLedgersHealth = function * () {
+  // TODO: check ledgers concurrently
+  yield Object.keys(this.ledgers).map((ledgerId) => this.ledgers[ledgerId].checkHealth())
+  this.ledgersHealth = { ledgersHealth: 'OK' }
```
clark800: Use `src/common/health.js`?

src/lib/ledgers/multiledger.js
```diff
@@ -7,32 +7,46 @@ function Multiledger (options) {
...
+  return Promise.resolve()
```
sentientwaffle: This line is unnecessary.

src/lib/route-broadcaster.js
gip: Is there a easy way to catch all the network-related exceptions?

lumberj: No, you'd have to do something like this: https://github.com/interledger/five-bells-connector/blob/9e277604758222b7ad30787598159e2b2855a4a0/src/controllers/notifications.js#L94

Unless all Network related exceptions derived from the same base class.

src/lib/ledgers/multiledger.js
```diff
@@ -1,38 +1,51 @@
...
 
 const FiveBellsLedger = require('./five-bells-ledger')
+const healthStatus = require('../../common/health.js')
 
 function Multiledger (options) {
   this.config = options.config
   this.log = options.log
   this.ledger_types = {undefined: FiveBellsLedger}
   this.ledgers = {}
+  this.buildLedgers()
```
lumberj: The previous build ledger didn't mutate the instance, it just returned a new ledger. So maybe, `this.ledgers = buildLedgers()` is closer to that.

src/lib/ledgers/multiledger.js
```diff
@@ -1,38 +1,51 @@
...
   })
 }
 
 Multiledger.prototype.getType = function (ledgerId) {
   return this.getLedger(ledgerId).constructor.TYPE
 }
 
+Multiledger.prototype.checkLedgersHealth = function * () {
+  yield Object.keys(this.ledgers).map((ledgerId) => this.ledgers[ledgerId].checkHealth())
+  this.ledgersHealth = { ledgersHealth: healthStatus.statusOk }
```
lumberj: Once this is set to statusOk the first time, it can never be set to notOk because checkHealth will retrying with backoff forever. Is there ever a point where we might want to timeout?

lumberj: What if, prior to checking ledger health, you reset it back to notOk?

src/lib/route-broadcaster.js
gip: Done. Catching only system errors, otherwise error is rethrown.

src/lib/ledgers/multiledger.js
```diff
@@ -1,38 +1,53 @@
...
 }
 
 Multiledger.prototype.addLedger = function (Ledger) {
   this.ledger_types[Ledger.TYPE] = Ledger
 }
 
 Multiledger.prototype.getLedger = function (ledgerId) {
-  return this.ledgers[ledgerId] ||
-        (this.ledgers[ledgerId] = this.buildLedger(ledgerId))
+  return this.ledgers[ledgerId]
```
clark800: This is a behavioral change, what is the reasoning for it?

gip: In the old implementation, `this.ledgers` was populated lazily when `getLedger()` was called. In the new one, this is done in the ctor since `checkLedgersHealth()` needs to be aware of all ledgers.

clark800:
LGTM

PR #157, opened 2016-05-26
WebSocket: Reconnect on disconnect

vhpoet:
Websocket was not reconnecting when one of the ledgers was going down, so all of the payments were getting expired after the ledger recovery.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "moment": "^2.10.2",
     "node-uuid": "^1.4.2",
     "passport-client-certificate": "^0.1.1",
+    "reconnect-core": "^1.2.0",
     "spdy": "^3.2.3",
     "uuid4": "^1.0.0",
     "ws": "^1.1.0"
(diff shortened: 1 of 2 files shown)


emschwartz:
Need to fix lint error:
image

emschwartz:
@vhpoet why was the ledger going down?

vhpoet:
@emschwartz I was just working with the ledger so had to restart sometimes, but I also noticed that socket connection was going down for no reason, and it was not recovering.

emschwartz:
Suggested trying https://github.com/juliangruber/reconnect-core instead of custom reconnect logic

vhpoet:
@emschwartz check it

emschwartz:
This will need to be modified further later but LGTM for now

MatthewPhinney:
> This will need to be modified further later but LGTM for now

@emschwartz how does this need to be modified?

emschwartz:
With this change it'll try to reconnect automatically, and it will presumably keep trying to reconnect. If it fails to reconnect after some amount of time it should probably stop processing payments completely. Especially in Universal mode, failure to receive notifications on one side could be exploitable.

MatthewPhinney:
Got it. Thanks for the explanation.

lumberj:
> Especially in Universal mode, failure to receive notifications on one side could be exploitable.

Even after it reconnects, it will never receive any notifications that were sent during the disconnection period

emschwartz:
Correct

PR #159, opened 2016-05-31
Refactor: Use ledger plugin framework

justmoon:
Builds on #152 so that PR should be merged first.

This implements the ledger plugin interface defined here for all interactions with Five Bells Ledger:

https://interledger.org/rfcs/0004-ledger-plugin-interface/

This will allow us to use the same ledger abstraction layer across -connector, -sender and -receiver. I'm hoping to refactor the basic ILP functionality into a module called `ilp-core` soon that both sender/receiver and connector use. This module will contain `RouteTables` and `RouteBuilder`, but not `RouteBroadcaster` ([CCP](https://interledger.org/rfcs/0010-connector-to-connector-protocol/)) and public-facing quoting functionality.

This also allows the creation of other ledger plugins, which is a prerequisite for supporting existing ledgers and making Interledger truly, well... inter-ledger.


src/lib/route-builder.js
```diff
@@ -83,24 +84,42 @@ class RouteBuilder {
...
+    }
+
+    const noteToSelf = {
+      source_transfer_ledger: sourceTransfer.ledger,
+      source_transfer_id: sourceTransfer.id
     }
 
     return _.omitBy({
-      id: nextHop.destinationLedger + '/transfers/' + uuid(),
+      // TODO: Use a real secret
```
emschwartz: should explain what this secret is used for (this is the kind of TODO everyone forgets what it was in there for)

src/lib/utils.js
```diff
@@ -16,6 +18,30 @@ function getPairs (arr) {
   ), [])
 }
 
+/**
+ * Deterministically generate a UUID from a secret and a public input.
+ *
+ * Uses HMAC-SHA-256 to generate a new UUID given a secret and a public input.
```
emschwartz: probably worth explaining what this is for

src/models/payments.js
```diff
@@ -1,60 +1,18 @@
...
-
-function * validateExpiry (payment, config) {
+function * validateExpiry (sourceTransfer, destinationTransfer, config) {
   // TODO tie the maxHoldTime to the fx rate
   // TODO bring all these loops into one to speed this up
-  const tester = yield testPaymentExpiry(config, payment)
+  const tester = yield testPaymentExpiry(config, sourceTransfer, destinationTransfer)
   tester.validateNotExpired()
   if (tester.isAtomic()) {
     tester.validateMaxHoldTime()
```
emschwartz: `tester.validateMaxHoldTime()` is called in both branches of the `if...else` so it should just be taken out and moved after  `tester.validateNotExpired()`

src/models/payments.js
```diff
@@ -64,178 +22,15 @@ function * validateExpiry (payment, config) {
   }
 }
 
-function amountFinder (ledger, creditOrDebit, config) {
-  const accountUri = config.getIn(['ledgerCredentials', ledger, 'account_uri'])
-  return (creditOrDebit.account === accountUri
-    ? new BigNumber(creditOrDebit.amount)
-    : new BigNumber(0))
+function * validate (sourceTransfer, destinationTransfer, config) {
```
emschwartz: what's the point of this function if it only calls `validateExpiry`?

test/notificationSpec.js
```diff
@@ -364,31 +365,6 @@ describe('Notifications', function () {
...
-          result: 'ignored',
-          ignoreReason: {
-            id: 'UnacceptableConditionsError',
-            message: 'Each of the source transfers\' execution conditions must match all of the destination transfers\' conditions'
-          }
-        })
-        .end()
-    })
-
     it.skip('should return a 422 if the two transfer conditions do not ' +
```
emschwartz: this can be removed too

src/ledgers/five-bells-ledger.js
```diff
@@ -172,6 +179,96 @@ class FiveBellsLedger extends EventEmitter {
...
+    let handled = false
+    for (let credit of fiveBellsTransfer.credits) {
+      if (credit.account === this.credentials.account_uri) {
+        handled = true
+
+        const transfer = lodash.omitBy({
+          id: fiveBellsTransfer.id.substring(fiveBellsTransfer.id.length - 36),
+          direction: 'incoming',
+          ledger: this.id,
+          // TODO: What if there are multiple debits?
```
emschwartz: should we just remove the multi-crebit feature from the f-b-ledger?

src/ledgers/five-bells-ledger.js
```diff
@@ -172,6 +179,96 @@ class FiveBellsLedger extends EventEmitter {
...
+        }
+      }
+    }
+
+    for (let debit of fiveBellsTransfer.debits) {
+      if (debit.account === this.credentials.account_uri) {
+        handled = true
+
+        // This connector only launches transfers with one credit, so there
+        // should never be more than one credit.
```
emschwartz: where is this checked?

src/lib/testPaymentExpiry.js
```diff
@@ -100,40 +99,12 @@ TransferTester.prototype.validateExpiryHoldTime = function (expiresAt) {
...
 }
 
-// We also need to check if we have enough time between the expiry
-// of the destination transfer with the latest expiry and the expiry of
-// the source transfer with the earliest expiry is greater than the
-// minMessageWindow.
-// This is done to ensure that we have enough time after the last
-// moment one of the destination transfers could happen (taking money out
-// of our account) to execute all of the source transfers
-TransferTester.prototype.validateMinMessageWindow = function () {
```
emschwartz: why is this being removed?

src/models/notifications.js
```diff
@@ -32,7 +32,12 @@ function verifySignature (notification, config) {
 
 function * processNotification (notification, ledgers, config) {
   if (notification.event === 'transfer.update') {
-    yield payments.updateTransfer(
+    const ledger = ledgers.getLedger(notification.resource.ledger)
```
emschwartz: should the `ledger` field be a property on the `notification` instead of the `resource` (to support non-5b-ledgers)?

src/models/payments.js
```diff
@@ -33,62 +24,28 @@ function * settle (sourceTransfer, destinationTransfer, config, ledgers) {
...
+  yield settle(sourceTransfer, destinationTransfer, config, ledgers)
 }
 
-function * updateDestinationTransfer (updatedTransfer, traderDebit, relatedResources) {
-  if (updatedTransfer.state !== 'executed') {
-    log.debug('Got notification about unknown or incomplete transfer: ' + updatedTransfer.id)
-    return
+function * processExecutionFulfillment (transfer, fulfillment) {
+  if (transfer.direction === 'outgoing') {
+    log.debug('Got notification about executed destination transfer')
```
emschwartz: log the transfer id?

src/ledgers/five-bells-ledger.js
```diff
@@ -183,7 +204,11 @@ class FiveBellsLedger extends EventEmitter2 {
     return null
   }
 
-  * fulfillCondition (transferID, conditionFulfillment) {
+  fulfillCondition (transferID, conditionFulfillment) {
+    return co.wrap(this._fulfillCondition).call(this, transferID, conditionFulfillment)
```
emschwartz: what's the point of using `co.wrap` for this and `getConnectors`? why not just change those functions to return promises?

emschwartz:
LGTM once tests pass

src/ledgers/five-bells-ledger.js
clark800: I was wondering this too, what was the use case for it?

src/models/payments.js
justmoon: Will do.

src/models/notifications.js
justmoon: That would be a breaking change. (This is a non-breaking PR.)

My guess is we'll probably just take the HTTP notifications out entirely. They are no longer used, but I kept them in because they are still used by the majority of our tests for the internal business logic. I added subscriptionSpec.js so that we have a new test suite that we can migrate those tests to (in future refactorings.)

src/lib/testPaymentExpiry.js
justmoon: This PR makes a change so that the connector chooses the destination expiry. Since the connector chooses it, we don't need to worry about it being valid. The only thing we need to make sure is that both expiries are still in the future, so that check is still in there.

src/ledgers/five-bells-ledger.js
justmoon: The ledger plugin interface only allows outgoing transfers with one debit. So if there is a transfer with more than one debit, it would have to come from some other local ledger client using the same account. So presumably it wouldn't have a noteToSelf or other markers of an ILP transfer, so it would be ignored.

I'll go back and double-check the logic to make sure we are not inadvertently processing such outgoing transfers.

justmoon: I should mention: Even if we are processing any such transfers - there isn't really anything bad that can happen since all we do in response to outgoing transfers is fulfill incoming ones. So the gun and the foot are pretty far apart here.

src/ledgers/five-bells-ledger.js
justmoon: Yeah, ILP would no longer used multi-crebit. I still think it's a good feature for a ledger to have and it could still be useful for us, so I don't see a strong reason to remove it, but we can chat about it.

It shouldn't be part of this PR, because it would be a breaking change. (And a fairly major one at that.)

test/notificationSpec.js
justmoon: Will do.

src/models/payments.js
justmoon: Oops, was going to take it out, but forgot. Will remove.

src/lib/utils.js
justmoon: Will do.

src/lib/route-builder.js
justmoon: Will do.

src/lib/testPaymentExpiry.js
emschwartz: Don't we also need to compare it to the packet expiry?

src/app.js
```diff
@@ -59,7 +59,6 @@ function listen (koaApp, config, ledgers) {
   // Start a coroutine that connects to the backend and
   // subscribes to all the ledgers in the background
   co(function * () {
-    yield ledgers.checkLedgersHealth()
```
MatthewPhinney: Why was this removed? I think @gip added it recently.

src/lib/testPaymentExpiry.js
justmoon: After discussion with @emschwartz and @bensharaf:

We don't have that information in the data format right now, but we may add that feature later. If we provide a way for the sender to optionally specify the minimum destinationExpiryDuration, connectors may optionally respect that information and avoid putting money on hold when they don't need to.

emschwartz: (for the record we may want to express the "destinationExpiryDuration" as two dates instead of a duration: `destinationPrepareBy` and `destinationExpiresAt` or something like that)

src/app.js
justmoon: We are now checking ledger health via the websocket status. Websocket do regular pings automatically to ensure the connection is still live, so it's a more reliable mechanism than checking the health endpoint once at startup.

src/models/subscriptions.js
```diff
@@ -18,18 +49,15 @@ function * subscribePairs (pairs, ledgersService, config) {
...
-      yield payments.updateTransfer(resource, relatedResources, ledgersService, config)
-    }).catch((err) => {
-      log.warn('error processing notification: ' + err)
-    })
-  })
+function * subscribeLedger (ledgerUri, ledgersService, config) {
+  log.info('subscribing to ' + ledgerUri)
+  const ledger = ledgersService.getLedger(ledgerUri)
+
+  ledger.connect()
```
lumberj: ledger.connect() returns a Promise, I think it should be returned (otherwise how do you catch an exception in `connect`). Also, does this need to be a generator function?

PR #161, opened 2016-06-08
[FIX] Fix demo routing (rounding); slippage=0

sentientwaffle:
Requires https://github.com/interledger/five-bells-routing/pull/4

Overestimate rates on all curves to account for rounding errors, so that payments don't fail. This should fix the issues w/ the demo.


src/lib/config.js
```diff
@@ -212,7 +212,8 @@ function getLocalConfig () {
   // The spread is added to every quoted rate
   const fxSpread = Number(Config.getEnv(envPrefix, 'FX_SPREAD')) || DEFAULT_FX_SPREAD
```
mwshortt: Could you do the same thing to `fxSpread` as you did for `slippage`? I'm running into an error now that this ignores what you set it `CONNECTOR_FX_SPREAD`, but if you change this to
```
const fxSpreadString = Config.getEnv(envPrefix, 'FX_SPREAD') 
const fxSpread = fxSpreadString ? Number(fxSpreadString) : DEFAULT_FX_SPREAD
```
it works as expected

sentientwaffle: Done; thanks for pointing it out.

mwshortt: Wonderful - thank you!

src/lib/route-builder.js
```diff
@@ -163,21 +174,26 @@ class RouteBuilder {
...
+    return scale ? (1 / Math.pow(10, scale)) : 0
+  }
 }
 
 function throwAssetsNotTradedError () {
   throw new AssetsNotTradedError('This connector does not support the given asset pair')
 }
 
-function validatePrecision (amount, precision, ledger) {
+function validatePrecision (amount, precision, sourceOrDestination) {
```
justmoon: Why is this a better name than `ledger`?

src/lib/route.js
```diff
@@ -0,0 +1,108 @@
+'use strict'
+
+const routing = require('five-bells-routing')
+const ROUTE_EXPIRY = 45 * 1000 // milliseconds
+
+class Route extends routing.Route {
```
justmoon: When you extend a class, give it a name that explains how it's different. Looking at the code I'm actually not at all sure what it is changing vs the base class.

It also duplicates a lot of code, presumably because you need to override methods that explicitly reference the base class in the original code. This should be fixed by using `new this.constructor` or `new this` instead of `new Route`.

justmoon: > This should be fixed by using new this.constructor or new this instead of new Route.

I mean this should be fixed in `five-bells-routing`.

src/lib/route-broadcaster.js
```diff
@@ -121,6 +124,26 @@ class RouteBroadcaster {
       ]
     }
   }
+
+  // Shift the graph down by a small amount so that precision rounding doesn't
+  // cause UnacceptableRateErrors.
+  * _shiftRoute (route) {
```
justmoon: Shifting the points down by a fixed amount should be in `five-bells-routing`.

But `_getScaleAdjustment` can stay here.

src/lib/route-builder.js
```diff
@@ -163,21 +174,26 @@ class RouteBuilder {
       return credentials.account_uri === funds.account
     })
   }
+
+  * _getScaleAdjustment (ledger) {
```
justmoon: Duplicate method - this also exists in RouteBroadcaster.

justmoon:
LGTM

PR #162, opened 2016-06-10
Explain parameter for /quote to get information about how amounts are computed

gip:
* If `explain` parameter is set to `true` during a quote, an additional info field will be included in the response
* The base rate will be included iff the quote is local
* Aggregating additional info when a multi hop payment needs to be defined further and is not yet implemented. No non-local quotes will not return additional info.
* More fields for fees, slippage, spread to be added at a later stage.

### DIFF
--- a/schemas/Quote.json
+++ b/schemas/Quote.json
@@ -28,7 +28,11 @@
       "description": "Time in seconds between proposed_at and expires_at. Set in quotes from payment systems but not valid in actual transfers",
       "$ref": "NonNegativeDuration.json"
     },
-    "destination_expiry_duration": { "$ref": "NonNegativeDuration.json" }
+    "destination_expiry_duration": { "$ref": "NonNegativeDuration.json" },
+    "additional_info": {
+      "description": "Additional information related to the quote",
+      "type": "object"
+    }
   },
   "required": [
     "source_ledger",
--- a/src/lib/routing-tables.js
+++ b/src/lib/routing-tables.js
@@ -23,7 +23,8 @@ class RoutingTables {
         (this.sources[localRoute.source_ledger] = new routing.RoutingTable())
       const route = new routing.Route(localRoute.points, {
         minMessageWindow: localRoute.min_message_window,
-        nextLedger: localRoute.destination_ledger
+        nextLedger: localRoute.destination_ledger,
+        additional_info: localRoute.additional_info
       })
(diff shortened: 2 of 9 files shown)


emschwartz:
What is this used for?

gip:
A quote / transfer object basically shows amounts. The additional info field is an explanatory field: it basically explains the way the amounts were computed. This is important for banks since they will want to know what the base rate (or slippage, or fee) was for instance. Recomputing the rate from the amounts may lead to a number that is close (but not strictly equal) to the initial base rate and this is unacceptable for some applications.

Due to time pressure this PR is a minimal implementation for local quote only. 
Note: I'm thinking to change the PR to make returning additional_info optional depending on a parameter.

emschwartz:
Got it, thanks. The schema should also be updated: https://github.com/interledger/five-bells-connector/blob/master/schemas/Quote.json

src/lib/routing-tables.js
```diff
@@ -23,7 +23,8 @@ class RoutingTables {
         (this.sources[localRoute.source_ledger] = new routing.RoutingTable())
       const route = new routing.Route(localRoute.points, {
         minMessageWindow: localRoute.min_message_window,
-        nextLedger: localRoute.destination_ledger
+        nextLedger: localRoute.destination_ledger,
+        additional_info: localRoute.additional_info
```
sentientwaffle: To follow the convention, on the `info` object it should be `additionalInfo`.

gip:
@justmoon Could you please have a look to this?

schemas/Quote.json
```diff
@@ -28,7 +28,21 @@
       "description": "Time in seconds between proposed_at and expires_at. Set in quotes from payment systems but not valid in actual transfers",
       "$ref": "NonNegativeDuration.json"
     },
-    "destination_expiry_duration": { "$ref": "NonNegativeDuration.json" }
+    "destination_expiry_duration": { "$ref": "NonNegativeDuration.json" },
+    "additional_info": {
```
justmoon: If you have a property called `additional_info` it should be schemaless. The point of having an `additional_info` is that it's not part of the schema and you can easily add stuff to it, but you should also not rely on it existing.

gip: Changed. The only constraint left is that `additional_info` has to be an object. And indeed, it may or may not exist.

justmoon:
LGTM

PR #165, opened 2016-06-14
Passing additional info from quoter backend

gip:


clark800:
LGTM after tests pass

PR #166, opened 2016-06-15
[FIX] Handle unsupported currency pair

gip:


lumberj:
@gip for these bug fixes, are you going to cherry pick them to master also?

gip:
Yes, most definitely. I added that to the ticket - in a way it's part of the DoD.

src/backends/ilp-quoter/index.js
```diff
@@ -30,9 +30,13 @@ class ILPQuoter {
 
   * putPair (uri) {
     const req = request({ method: 'PUT', uri, json: true })
-    const res = yield req
-    if (res.statusCode >= 400) {
-      throw new UnsupportedPairError(req.uri)
+    try {
+      const res = yield req
+      if (res.statusCode >= 400) {
```
lumberj: I think what you want is something like this:

```
const res = yield req
if (res.statusCode >= 400) {
  log.error('Unsupported Pair: ' + uri + ' ' + err.message)
  throw new UnsupportedPairError(req.uri)
}
```

Otherwise, the error will not bubble up to the request handler

lumberj:
Can you describe the bug this is supposed to fix? It's not obvious

src/backends/ilp-quoter/index.js
gip: The point of this fix is not to throw though -- otherwise the call to `* connect ()` will fail without having enabled all the pairs. Please refer to original ticket.

lumberj: Can you add a unit test case for this? 

> Please refer to original ticket.

What ticket?

gip:
@lumberj Ok, please see ticket #590. When `* connect ()` is called, the connector will do a handshake with the quoter, informing the quoter which currency pairs the connector deals with. Before this fix, if the quoter would return 404 (currency pairs not supported) then `* connect ()` would just fail at once even though the handshake wasn't completed.

test/ilpQuoterSpec.js
```diff
@@ -54,7 +53,7 @@ describe('ILPQuoter', function () {
                       .put('/pair/EUR/USD').reply(200)
                       .put('/pair/XRP/USD').reply(400)
```
lumberj: So the connector returns a 400 for an unsupported pair?

src/backends/ilp-quoter/index.js
clark800: We should treat 4xx and 5xx differently as Alan suggests, otherwise you will get very confusing errors when the server is down

clark800: In fact, it's probably best to check just for the exact error code the quoter is supposed to generate for this case and pass the error on for all other cases

gip:
@lumberj If you try to get a quote for an unsupported pair it will fail obviously. 
Basically we have to decide what the error mode for the connector when 1+ currency pair is not supported by the quoter. We could either (1) fail and stop the connector or (2) continue and work with the currency pairs that were validated by the quoter. The behavior of the connector before this fix is (as described in #590 ticket) in between and inconsistent - some valid pairs will work, some other won't. 
This PR implements (2) but we can do (1) if that seems better?

lumberj:
It makes sense to me for the Connector to fail if it supports pairs that the backend does not.

Thoughts from @MatthewPhinney or @clark800 ?

clark800:
> It makes sense to me for the Connector to fail if it supports pairs that the backend does not.

Have any justifications? Here's mine: I would assume that this would almost certainly be due to a misconfiguration and it would be better to fail immediately than let the user run the service for a while before someone discovers that it was misconfigured.

MatthewPhinney:
The connector supporting pairs that the quoter doesn't support is a configuration issue, right? It probably makes sense for the connector to fail if it's started with pairs that the backend doesn't support. Otherwise, someone hitting the connector `pairs` endpoint will get back a list of pairs, some valid and some not. This seems undesirable to me.

gip:
To follow up on that: what should be the actual behavior? After an error message is printed, should the connector exit or keep running with a failing health check?

MatthewPhinney:
My instinct would be to exit with an error, since the issue is caused by an incorrect configuration. The only way to fix it is to restart the connector with the correct config, right?

lumberj:
> To follow up on that: what should be the actual behavior? After an error message is printed, should the connector exit or keep running with a failing health check?

Exiting sounds right, otherwise it will run with a faulty configuration and end up with runtime errors

PR #168, opened 2016-06-20
Refactor: remove ledger-specific code

justmoon:
Move ledger specific code to a separate ledger plugin (`ilp-plugin-bells`).

There is only one piece of ledger-specific code remaining after this PR: The info cache still assumes a far-away ledger is a Five Bells ledger. This should be changed, but I couldn't think of a way to do it as a non-breaking change.

Changes the `account_uri` field in `LEDGER_CREDENTIALS` to just `account` to be consistent with the names of other fields. (We don't normally mention the format in the field name, e.g. `amount_decimal` or `ledger_uri`.) Accepts old format with a deprecation warning.


src/lib/multiledger.js
```diff
@@ -1,21 +1,16 @@
 'use strict'
 
 const _ = require('lodash')
-const FiveBellsLedger = require('../ledgers/five-bells-ledger')
 const healthStatus = require('../common/health.js')
 
 function Multiledger (options) {
```
emschwartz: Should the connector use `ilp-core` instead of this multiledger abstraction, or is that for another PR?

emschwartz:
LGTM once tests pass

README.md
```diff
@@ -109,17 +109,6 @@ ex.
 * `CONNECTOR_TLS_CRL` (default: none) the path to the server certificate revokation list file. Optional if using HTTPS.
 * `CONNECTOR_TLS_CA` (default: none) the path to a trusted certificate to be used in addition to using the [default list](https://github.com/nodejs/node/blob/v4.3.0/src/node_root_certs.h). Optional if using HTTPS.
 
-#### Auto-funding
```
emschwartz: Will the demo just make the calls to fund the connector directly?

src/lib/multiledger.js
justmoon: Yeah, switching to `ilp-core` would be the next step.

README.md
justmoon: That's correct.

PR #173, opened 2016-06-22
Fatal error when a currency pair is not supported by the backend

gip:


MatthewPhinney:
Is this related to #166

src/app.js
```diff
@@ -59,7 +59,11 @@ function listen (koaApp, config, ledgers) {
   // Start a coroutine that connects to the backend and
   // subscribes to all the ledgers in the background
   co(function * () {
-    yield backend.connect()
+    const success = yield backend.connect()
+    if (!success) {
+      log('app').error('backend initialization has failed - the connector will now exit')
+      process.exit(1)
```
gip: Now `* connect ()` returns a boolean and the component exits when this is not successful. This happens asynchronously after the server has been started and routes are available. This should be ok as at that time the health check will show that the component isn't ready.

gip:
@MatthewPhinney Yes new PR related to same issue

src/backends/ilp-quoter/index.js
```diff
@@ -42,9 +42,10 @@ class ILPQuoter {
    */
   * connect () {
     const uris = _.uniq(this.currencyPairs.map((pair) => this.backendUri + '/pair/' + pair[0] + '/' + pair[1]))
-    yield uris.map((uri) => this.putPair(uri))
-    this.backendStatus = healthStatus.statusOk
-    return
+    const results = yield uris.map((uri) => this.putPair(uri))
```
MatthewPhinney: So `results` is an array of booleans? I didn't realize you can `yield` over a map in which the callback function returns a generator.

src/backends/ilp-quoter/index.js
```diff
@@ -32,8 +30,10 @@ class ILPQuoter {
     const req = request({ method: 'PUT', uri, json: true })
     const res = yield req
     if (res.statusCode >= 400) {
-      throw new UnsupportedPairError(req.uri)
+      log.error('Unsupported pairs, the quoter reported an error on PUT ' + uri)
```
MatthewPhinney: Is there any way to distinguish between different types of quoter errors? Maybe log the res body to give the user more info.

src/app.js
lumberj: Why return a boolean instead of throwing an exception?

gip: Because I wanted all the calls to complete so that we can show an error message for *all* missing currency pairs (as apposed to only the first one -- that would happen if we throw an exception as soon as one call fails.

lumberj: You can change the code in the backend to handle that. Specifically, [this part](https://github.com/interledger/five-bells-connector/pull/173/files/26409ef3c1ea9d91d82f9340ef6b051d0e46fe68#diff-179f9c6e4bfdb0a397cc134cd45e1b48R46)

gip: What is the issue with returning a boolean? Meaning is clear. I don't see why returning undefined / exception is better?

src/backends/ilp-quoter/index.js
gip: Changed

src/app.js
MatthewPhinney: The main difference I see between throwing an exception and returning a value is that with an exception, if the calling method forgets to catch the exception, it will bubble up to the top of the call stack, but if it forgets to check the response, nothing will happen. But as long as the app checks the response from `connect`, I don't see an issue with this.

src/backends/ilp-quoter/index.js
gip: Yes it is an array of booleans.

src/app.js
gip: In synchronous code I would agree. However this is part of a coroutine and is asynchronous. The former implementation was actually throwing an exception if a pair wasn't correct. The only effect was to make a promise fail, silently. That was basically the bug. So making sure `connect` is typed makes sense here I think.

MatthewPhinney:
LGTM

gip:
@lumberj Changed

src/backends/ilp-quoter/index.js
```diff
@@ -42,9 +50,13 @@ class ILPQuoter {
    */
   * connect () {
     const uris = _.uniq(this.currencyPairs.map((pair) => this.backendUri + '/pair/' + pair[0] + '/' + pair[1]))
-    yield uris.map((uri) => this.putPair(uri))
-    this.backendStatus = healthStatus.statusOk
-    return
+    const results = yield uris.map((uri) => this.putPair(uri))
+    const success = _.every(results, (result) => result.success)
+    if (!success) {
+      const message = _.reduce(results, (msg, result) => result.success ? msg : msg + result.errorMessage + ' ', '')
```
lumberj: This is fine, but there's another helper in lodash:

```javascript
const message = _.join(_.map(_.reject(results, 'success'), 'errorMessage'), ' ')
```

lumberj:
LGTM

PR #175, opened 2016-06-23
Fix: Support ILP-Plugin-Virtual

bensharaf:
Changes multiledger and testPaymentExpiry so that ILP-Plugin-Virtual is able to be used in the connector. This includes adding a sqlite-based store in src/lib for persistence in ILP-Plugin-Virtual. ILP-Plugin-Virtual is also added as a dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "five-bells-routing": "~3.1.0",
     "five-bells-shared": "^16.1.0",
     "ilp-plugin-bells": "^2.1.1",
+    "ilp-plugin-virtual": "^1.2.0",
     "koa": "^1.0.0",
     "koa-compress": "^1.0.6",
     "koa-cors": "0.0.16",
@@ -58,6 +59,7 @@
     "passport-client-certificate": "^0.1.1",
     "reconnect-core": "^1.2.0",
     "spdy": "^3.2.3",
+    "sqlite3": "^3.1.4",
     "uuid4": "^1.0.0",
     "ws": "^1.1.0"
   },
--- a/src/lib/sqliteStore.js
+++ b/src/lib/sqliteStore.js
@@ -1,0 +1,43 @@
+'use strict'
+
+const sqlite3 = require('sqlite3').verbose()
+
+function newSqliteStore (address) {
+  address = address || ':memory:'
+
+  const db = new sqlite3.Database(address)
+  db.run('CREATE TABLE IF NOT EXISTS store (key TEXT, value TEXT)', () => {
+    db.run('CREATE UNIQUE INDEX IF NOT EXISTS id ON store (key)')
+  })
+
+  const put = function (k, v) {
+    return new Promise((resolve) => {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -72,6 +73,7 @@
     "eslint-plugin-promise": "^1.1.0",
     "eslint-plugin-standard": "^1.3.2",
     "five-bells-integration-test": "^2.0.0",
+    "ilp-plugin-virtual": "^1.1.0",
```
emschwartz: This shouldn't be a devDependency

src/lib/info-cache.js
```diff
@@ -17,8 +15,10 @@ class InfoCache {
       // Default to Five Bells Ledger
       // TODO: This hack is necessary to fetch the precision information for
       // non-adjacent ledgers. How can we handle this case properly?
+      const FiveBellsLedger = require('ilp-plugin-bells')
```
MatthewPhinney: We generally put `require`s at the top of the file.

src/lib/sqliteStore.js
```diff
@@ -0,0 +1,43 @@
+'use strict'
+
+const sqlite3 = require('sqlite3').verbose()
+
+function newSqliteStore (address) {
+  address = address || ':memory:'
+
+  let db = new sqlite3.Database(address)
```
justmoon: Should be `const`, not `let` because you're not going to reassign a different value later.

src/lib/sqliteStore.js
```diff
@@ -0,0 +1,43 @@
...
+
+  let del = function (k) {
+    return new Promise((resolve) => {
+      db.run('DELETE FROM store WHERE (key == ?)', k, () => {
+        resolve()
+      })
+    })
+  }
+
+  return { get: get, put: put, del: del }
```
justmoon: Use the shorthand:

```js
return { get, put, del }
```

src/lib/info-cache.js
bensharaf: That change is from 'remove ledger specific code,' so it should probably be addressed in pull request 168. I'm going to rebase my commit on top of that once it's merged.

src/lib/sqliteStore.js
bensharaf: Circle CI build fails when I change it to that version

bensharaf:
Fixed CI

src/lib/testPaymentExpiry.js
```diff
@@ -87,7 +87,7 @@ TransferTester.prototype.validateMaxHoldTime = function () {
 
   if (expiresAt) {
     this.validateExpiryHoldTime(expiresAt)
-  } else {
+  } else if (this.destinationTransfer.executionCondition) {
```
sentientwaffle: I don't think this change is needed (see https://github.com/interledger/five-bells-connector/pull/180).

justmoon:
LGTM

PR #177, opened 2016-06-27
[FIX] Connect to ledgers on startup

MatthewPhinney:


src/app.js
```diff
@@ -65,6 +65,11 @@ function listen (koaApp, config, ledgers) {
       log('app').error(error.message)
       process.exit(1)
     }
+    const ledgerUirs = Object.keys(ledgers.getLedgers())
+    for (let i = 0; i < ledgerUirs.length; i++) {
```
clark800: This is a "low-level" way of iterating over an array; would be better to use "map"

clark800:
What was it doing before? What is the reason for this change?

MatthewPhinney:
> What was it doing before? What is the reason for this change?

For some reason the code that checks the ledger health was removed in this refactor: https://github.com/interledger/five-bells-connector/commit/deb1dcdc05d229e8e086b94750347d802a886dc7#diff-bd9c9dcd314f2d7df52935b3a6a4d504L62.

src/app.js
MatthewPhinney: Yeah, I also prefer a more functional approach, but I was having trouble getting the generators to work with the callback for `map`. I could do something like this:
```
    for (let ledgerUri of Object.keys(ledgers.getLedgers())) {
      const ledger = ledgers.getLedger(ledgerUri)
      yield ledger.connect()
    }
```
but I'm open to suggestions.

clark800: I think `co` allows you to `yield` on an array of generator functions

clark800:
> For some reason the code that checks the ledger health was removed in this refactor

I don't understand the connection, can you clarify?

src/app.js
emschwartz: You might want to use `Promise.all()` instead of doing it sequentially

clark800: `yield`ing on an array also does parallel execution, I'm all for moving away from `co` though if everyone agrees

MatthewPhinney:
The connector was checking that it was connected to ledgers before trying to broadcast routes. This logic was removed in commit https://github.com/interledger/five-bells-connector/commit/deb1dcdc05d229e8e086b94750347d802a886dc7#diff-bd9c9dcd314f2d7df52935b3a6a4d504L62. This wasn't tested with ILP server until today, when I noticed an `ECONNREFUSED` in the connector when trying to crawl the ledgers. Adding back the logic to ensure we are connected to the ledgers fixed the issue in ILP server.

MatthewPhinney:
OK, to clarify, it looks like the logic to connect to the ledgers is inside the `subscribePairs` method. We need to ensure this method is called (so that we are connected to the ledgers) before calling `routeBroadcaster.start`

clark800:
LGTM

emschwartz:
Is `connect` being called twice now?

MatthewPhinney:
> Is connect being called twice now?

I don't believe so ... I removed the code in my PR that called connect.

PR #178, opened 2016-06-27
Allow FX_SPREAD and SLIPPAGE to be set to 0

mwshortt:


src/lib/config.js
```diff
@@ -206,9 +206,11 @@ function getLocalConfig () {
   expiry.maxHoldTime = +Config.getEnv(envPrefix, 'MAX_HOLD_TIME') || DEFAULT_MAX_HOLD_TIME
 
   // The spread is added to every quoted rate
-  const fxSpread = Number(Config.getEnv(envPrefix, 'FX_SPREAD')) || DEFAULT_FX_SPREAD
+  const fxSpreadString = Config.getEnv(envPrefix, 'FX_SPREAD')
+  const fxSpread = fxSpreadString ? +fxSpreadString : DEFAULT_FX_SPREAD
```
lumberj: Does `+string` behave differently than `Number(string)` for casting? If not, `Number()` show intent better

gip:
Ideally a unit test on `RouteBuilder` to test zero slippage could be added. But since it's a on short term branch that's ok. 
LGTM.

src/lib/config.js
mwshortt: I took this from the master branch: https://github.com/interledger/five-bells-connector/blob/master/src/lib/config.js#L214-L218. Basically manually cherrypicking. I am happy to change this to `Number()` but it would cause a conflict if we merge this to `master`.

lumberj: Oh let's leave that then! Thanks for clarifying

PR #182, opened 2016-07-05
[FIX] ledger status should be an object

MatthewPhinney:


MatthewPhinney:
@gip I think you originally added this code. The tests don't check the `/health` response body; the endpoint was erroneously reporting that the status was `WAITING`, when it was, in fact, `OK`.  Health endpoint now returns:
```
{
  "backendStatus": "OK",
  "ledgersStatus": "OK",
  "status": "OK"
}
```
If we mock up the ledger connection during the tests, we could check the health response body. Let me know if you think that's worthwhile.

src/lib/multiledger.js
```diff
@@ -41,7 +41,11 @@ Multiledger.prototype.getType = function (ledgerId) {
 }
 
 Multiledger.prototype.getStatus = function () {
-  return _.every(this.ledgers, (ledger) => ledger.isConnected())
+  return _.every(this.ledgers, (ledger) => ledger.isConnected()) ? {
```
gip: Couldn't this be also written:
```js
  return { 
    ledgersStatus: _.every(this.ledgers, (ledger) => ledger.isConnected()) ? healthStatus.statusOk : healthStatus.statusNotOk
  }
```
Sounds more natural not to duplicate the object maybe?

MatthewPhinney: Sure, that works.

gip:
LGTM

PR #183, opened 2016-07-06
feature: send ledger precision and scale to quoter when getting quote

lumberj:
feature: send ledger precision and scale to quoter when getting quote

The connector should send precision of the source or destination ledger
based on the quote value that needs to be computed (and rounded).

- If quote specifies source amount, then send the precision of the
  receiving ledger
- If quote specifies receive amount, then send the precision of the source ledger

### DIFF
--- a/src/backends/ilp-quoter/index.js
+++ b/src/backends/ilp-quoter/index.js
@@ -26,6 +26,7 @@ class ILPQuoter {
                                                                   p[1].slice(0, 3)])
     this.backendUri = opts.backendUri
     this.backendStatus = healthStatus.statusNotOk
+    this.infoCache = opts.infoCache
   }
 
   * putPair (uri) {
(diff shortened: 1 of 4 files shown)


sentientwaffle:
`RouteBuilder` already rounds quotes. Why does the backend need to round as well?

lumberj:
The backend can only provide so many decimal places when returning a quote. It's best if it uses the same precision settings as the source or destination ledger

src/backends/ilp-quoter/index.js
```diff
@@ -80,22 +81,32 @@ class ILPQuoter {
...
     if (params.source_amount) {
       amount = params.source_amount
       type = 'source'
-    }
-    if (params.destination_amount) {
+      ledger = params.source_ledger
+    } else if (params.destination_amount) {
       amount = params.destination_amount
       type = 'destination'
+      ledger = params.destination_ledger
```
clark800: Is this backwards? If you know the destination amount, then you want to round to the precision of the source ledger right?

lumberj: The ticket specifically says what I wrote in the commit message, 

> If quote is by source amount, then use precision of the sending ledger
> If quote is by receive amount, then use precision of the receiving ledger

clark800: Does that make sense to you? It doesn't make sense to me.

lumberj: You're right. It seems backwards

lumberj:
TODO:

- Use source ledger precision if it's a destination amount quote (calculated source amount needs to be rounded) and vice versa for source amount quote

lumberj:
Updated

clark800:
LGTM

PR #184, opened 2016-07-07
Allow specification of destination transfer id

mwshortt:


MatthewPhinney:
LGTM although it would be good if @stefan or @emschwartz signed off on this, as well.

emschwartz:
I'm not sure this should be standard connector functionality. What's the use case for this?

Also, there is an argument for why you would actually _not_ want earlier hops in a payment to know a later hop's transfer ID. If I'm an earlier connector, I can make a later connector seem unreliable because I can put in another transfer on that ledger with the same ID and cause that payment to fail. See the note [here](https://github.com/interledger/five-bells-connector/blob/2dffa0063d48e58ddd9b154ac37d45f334caa1fd/src/lib/route-builder.js#L131-L141)

MatthewPhinney:
@mwshortt is there a jira ticket for this? We can move the discussion there.

mwshortt:
@MatthewPhinney Not yet, I moved the discussion to Slack with @emschwartz yesterday and hope @justmoon can review this at some point today.

mwshortt:
Closed because a better solutions was implemented here: https://github.com/interledger/five-bells-ledger/pull/285 (this endpoint removes the need to keep track of the destination transfer id)

PR #188, opened 2016-07-11
fix: subscriptions.subscribePairs never yields

lumberj:
It needs to be in its own coroutine


lumberj:
Closing. Looking into comments from @sentientwaffle

PR #189, opened 2016-07-12
[TEST] use MockPlugin for tests

sentientwaffle:


src/lib/testPaymentExpiry.js
```diff
@@ -60,8 +60,7 @@ TransferTester.prototype.isAtomic = function () {
 TransferTester.prototype.validateNotExpired = function () {
   {
     const expiresAt = this.getExpiry(this.sourceTransfer)
-    if (expiresAt && this.sourceTransfer.state !== 'executed' &&
-      moment(expiresAt, moment.ISO_8601).isBefore(moment())) {
+    if (expiresAt && moment(expiresAt, moment.ISO_8601).isBefore(moment())) {
```
emschwartz: What was this change about? If the transfer has already executed why should it throw an error?

sentientwaffle: `this.sourceTransfer` isn't a five-bells transfer, so it doesn't even have a "state" property.

test/notificationSpec.js
```diff
@@ -321,305 +295,41 @@ describe('Notifications', function () {
...
+      }, {
+        label: 'should return 200 if the payment is from an unknown source ledger',
+        error: 'AssetsNotTradedError'
+      }, {
+        label: 'should return a 200 if the source transfer is expired',
+        error: 'UnacceptableExpiryError'
+      }
+    ].forEach(function (data) {
+      it(data.label, function * () {
+        this.ledgers.getLedger(this.notificationSourceTransferPrepared.resource.ledger)
```
emschwartz: I think it's more readable to spell each of these out rather than getting fancy...

test/notificationSpec.js
```diff
@@ -643,174 +353,11 @@ describe('Notifications', function () {
         .expect(200)
         .end()
     })
-
-    it('should execute a payment where the source transfer\'s expires_at date has passed if the transfer was executed before it expired', function * () {
```
emschwartz: This is a strange case, but shouldn't we support it?

test/notificationSpec.js
```diff
@@ -643,174 +353,11 @@ describe('Notifications', function () {
...
-          .send(this.notificationSourceTransferAtomic_TwoCases)
-          .expect(200)
-          .end()
-
-        nockRegTarget1.done()
-        nockRegTarget2.done()
-      })
-    })
-
-    describe('optimistic mode', function () {
```
emschwartz: Why is this being removed?

test/notificationSpec.js
sentientwaffle: Ok, I broke them up

test/notificationSpec.js
sentientwaffle: Moved/adapted to `test/paymentsSpec.js`

emschwartz:
LGTM

PR #190, opened 2016-07-12
Test for additional_info field

gip:
Added a test for backend `ilpQuoterSpec.js`. A test already exists in `routeBuilderSpec.js`.

Testing the actual `/quote` endpoint would involve instantiating the app with a mock ilpQuoter backend which goes beyond unit testing I guess. Thoughts?


MatthewPhinney:
LGTM

PR #193, opened 2016-07-18
fix: handle fulfillment reject instead of returning string

bensharaf:
This code currently expects `fulfillCondition` to resolve to a string, although this is not specified in the LedgerPlugin interface. Instead, the connector should give an error on a rejected promise.


justmoon:
Add test, otherwise LGTM.

PR #194, opened 2016-07-19
Feature flag for route shifting (on by default)

gip:
Feature flag for curve shifting. Some clients do not (yet) work with that feature. This flag allows to disable it until this is fixed. Should not impact other use cases since the feature is on by default.

### DIFF
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -17,6 +17,7 @@ const DEFAULT_SLIPPAGE = 0.001 // 0.1%
 const DEFAULT_ROUTE_BROADCAST_INTERVAL = 30 * 1000 // milliseconds
 const DEFAULT_ROUTE_CLEANUP_INTERVAL = 1000 // milliseconds
 const DEFAULT_ROUTE_EXPIRY = 45 * 1000 // milliseconds
+const DEFAULT_ROUTE_SHIFT = true
 
 function isRunningTests () {
   return (
@@ -192,6 +193,8 @@ function getLocalConfig () {
     Number(Config.getEnv(envPrefix, 'ROUTE_CLEANUP_INTERVAL')) || DEFAULT_ROUTE_CLEANUP_INTERVAL
   const routeExpiry =
     Number(Config.getEnv(envPrefix, 'ROUTE_EXPIRY')) || DEFAULT_ROUTE_EXPIRY
+  const routeShift =
+    Config.castBool(Config.getEnv(envPrefix, 'ROUTE_SHIFT'), DEFAULT_ROUTE_SHIFT)
 
   // Credentials should be specified as a map of the form
   // {
@@ -230,7 +233,8 @@ function getLocalConfig () {
     notifications,
     routeBroadcastInterval,
     routeCleanupInterval,
-    routeExpiry
+    routeExpiry,
+    routeShift
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/lib/config.js
```diff
@@ -17,6 +17,7 @@ const DEFAULT_SLIPPAGE = 0.001 // 0.1%
 const DEFAULT_ROUTE_BROADCAST_INTERVAL = 30 * 1000 // milliseconds
 const DEFAULT_ROUTE_CLEANUP_INTERVAL = 1000 // milliseconds
 const DEFAULT_ROUTE_EXPIRY = 45 * 1000 // milliseconds
+const DEFAULT_ROUTE_SHIFT = true
```
clark800: Title says "curve shifting", code says "route shift". What is actually shifting?

src/lib/config.js
```diff
@@ -192,6 +193,8 @@ function getLocalConfig () {
     Number(Config.getEnv(envPrefix, 'ROUTE_CLEANUP_INTERVAL')) || DEFAULT_ROUTE_CLEANUP_INTERVAL
   const routeExpiry =
     Number(Config.getEnv(envPrefix, 'ROUTE_EXPIRY')) || DEFAULT_ROUTE_EXPIRY
+  const routeShift =
+    Config.getEnv(envPrefix, 'ROUTE_SHIFT') ? Boolean(Config.getEnv(envPrefix, 'ROUTE_SHIFT')) : DEFAULT_ROUTE_SHIFT
```
clark800: Isn't this what the `castBool` method is for?

emschwartz:
@gip which clients do you mean? I think @sentientwaffle implemented this to fix a bug that caused payments to fail because of connector rounding. If I'm not mistaken about that it seems better to just fix the clients than turn this off and re-introduce that other issue.

sentientwaffle:
@gip disabling shifting will break multi-hop routing -- it might not be obvious that it's happening, though, because of the slippage, which grants some leeway.

src/lib/config.js
gip: A route is represented in this implementation by a curve, hence the confusion. I've changed the title.

src/lib/config.js
gip: Changed

gip:
@emschwartz @sentientwaffle I've added a feature flag so that our work can continue until we understand how to fix the issue. The current implementation produces wrong results with some local payments. I've sent @sentientwaffle with an explanation of what is wrong in that case. I'll create a ticket and copy you on this. 

I looked at the problem to try to fix it. It looks like in a 1 hop payment (from ledger 0 to ledger 2, going through ledger 1), the maximum loss for the sender/receiver (for a destination/source payment respectively) will be at most:
`maxLoss = 10^(-p0) + 2 * 10^(-p1)/r01 + 10^(-p2)/(r01 * r12)` 
where p0 , p1, p2 are the decimal precision for the ledgers 
     and  r02 and r12 are the exchange rate between the ledgers
However I do not understand how the curve shifting solves that problem for all cases. Do we have any documentation on this?

gip:
Also, the route shifting is enabled by default, so that shouldn't impact you guys.

emschwartz:
LFineTM for now. This sounds like a bug in the logic that should really be fixed instead of turned off but it sounds like it'll take a little bit of time to figure out what needs to be changed.

gip:
I agree, should be fixed instead of disabled. It's a tactical move - this PR aims at unblocking current development. The flag will be removed altogether when the bug is fixed.

PR #195, opened 2016-07-20
feat: BREAKING pass through ILP packet

emschwartz:
the connector was stripping the ilp header before sending the last
transfer. the full ilp packet is useful for the recipient to check the
details of the incoming transfer against what was supposed to happen
(i.e. what was in the packet).

this changes the behavior to pass
through the whole ILP packet. this is also useful for when we switch to
binary, as it is arguably simpler for the connector to just forward on
the packet rather than stripping particular fields before sending it to
the receiver. if we do want to strip fields before handing the packet
off to the actual end user, the client libraries can do that.

requires https://github.com/interledger/five-bells-receiver/pull/4 for integration tests to pass.

### DIFF
--- a/src/lib/route-builder.js
+++ b/src/lib/route-builder.js
@@ -143,7 +143,7 @@ class RouteBuilder {
       direction: 'outgoing',
       account: nextHop.destinationCreditAccount,
       amount: nextHop.destinationAmount,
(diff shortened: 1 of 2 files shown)


justmoon:
Mark as breaking change (for now we're doing this by putting `BREAKING` after the type of change, e.g. `feat(foo): BREAKING add coffee maker`)

Otherwise LGTM.

PR #196, opened 2016-07-20
feature: Improve logging in Connector

lumberj:
Related: https://github.com/interledger/five-bells-ledger/pull/294

Motivation:
To provide structured, machine-parseable logging for production
deployments.

The bunyan logger will by default log in JSON in one line:

```
 {"name":"koa","hostname":"alan-lm","pid":39269,"req_id":"bbf85791-9169-43bd-a679-4214c8b9cea7","level":30,"req":{"method":"GET","url":"/accounts/test","headers":{"host":"localhost:3001","accept":"application/json","connection":"close"},"remoteAddress":"::ffff:127.0.0.1","remotePort":60024},"res":{"statusCode":200,"header":"HTTP/1.1 200 OK\r\nAccess-Control-Allow-Origin: *\r\nAccess-Control-Expose-Headers: link\r\nAccess-Control-Allow-Methods: GET,HEAD,PUT,POST,DELETE\r\nContent-Type: application/json; charset=utf-8\r\nContent-Length: 162\r\nDate: Tue, 12 Jul 2016 23:49:22 GMT\r\nConnection: close\r\n\r\n"},"duration":9,"msg":"  --> GET /accounts/test 200 9ms","time":"2016-07-12T23:49:22.811Z","v":0}
 ```

This can be parsed for human-readability using a JSON parser such as [`jq`](https://stedolan.github.io/jq/) or the [bunyan CLI](https://github.com/trentm/node-bunyan#cli-usage)

In this commit, `npm start` will by default pipe the JSON output to the
`bunyan` CLI. Example logging looks like:

```
 [2016-07-12T23:49:22.978Z]  INFO: connector/39269 on alan-lm:   <-- GET /connectors (req_id=0053ab9c-de7f-4782-a55f-a33ca6042f58, req.remoteAddress=::ffff:127.0.0.1, req.remotePort=60037)
    GET /connectors HTTP/1.1
    host: localhost:3001
    accept: application/json
    connection: close
[2016-07-12T23:49:22.982Z]  INFO: connector/39269 on alan-lm:   --> GET /connectors 200 4ms (req_id=0053ab9c-de7f-4782-a55f-a33ca6042f58, duration=4, req.remoteAddress=::ffff:127.0.0.1, req.remotePort=60037)
    GET /connectors HTTP/1.1
    host: localhost:3001
    accept: application/json
    connection: close
    --
    HTTP/1.1 200 OK
    HTTP/1.1 200 OK
...
```

Additionally, the request logger will log the `X-Request-ID` header and
store the result in the Koa context as `reqId` (e.g., `this.reqId`).
This is helpful for tracing requests across components (e.g., ledger ->
connector, connector -> ledger, client -> connector -> ledger)
See:
https://github.com/koajs/bunyan-logger#koabunyanloggerrequestidcontextopts

log level can be configured using `CONNECTOR_LOG_LEVEL` environment
variable. Default is `debug` in test, and `info` otherwise. `trace`
level will log the request and response bodies

### CODE REVIEW

File: src/app.js [hunk 1]
```diff
@@ -119,8 +117,29 @@ function createApp (config, ledgers, backend, routeBuilder, routeBroadcaster, ro
...
+      return {
+        duration: fields.duration,
+        status: this.status,
+        headers: this.headers,
+        body: isTrace ? this.body : undefined
+      }
+    }
+  }))
+  koaApp.use(errorHandler({log: log.create('error-handler')}))
+  koaApp.on('error', function () {})
```
clark800: Why was this added?

lumberj: Otherwise, the app emits an error and the stack trace is logged directly to the console (it's already logged by the logger).

See: https://github.com/koajs/bunyan-logger#suppressing-default-error-stack-traces

lumberj: It should also be added to the ledger

clark800: If we were showing them before, then I think we should keep showing them because we only show stack traces when there is a problem that should never happen. For things that should never happen I think its best to show it everywhere because when you are developing it will waste time if the error is in the log and you aren't looking at the log at that moment.

lumberj: We are still showing them. They are logged here https://github.com/interledger/five-bells-shared/blob/master/middlewares/error-handler.js#L15

clark800: Does that go to stderr? I assume not or else we should have been seeing duplicate stack traces before.

lumberj: Yes. It goes to stderr. We can leave it, but it will show up in the test output

lumberj: Also, it's not very helpful, because those things get logged without timestamps or any other contextual information. Just a stacktrace

clark800: Ok, if unexpected error stack traces go to stderr then it should be good

clark800:
LGTM

PR #197, opened 2016-07-21
[BREAKING] ILP Addressing; use ilp-core

sentientwaffle:
Depends on:

* https://github.com/interledger/js-ilp-plugin-bells/pull/16
* https://github.com/interledger/js-ilp-core/pull/25
* https://github.com/interledger/five-bells-shared/pull/150
* https://github.com/interledger/five-bells-integration-test/pull/26


src/models/quote.js
```diff
@@ -1,38 +1,29 @@
...
-    uri: account,
-    json: true
-  })
-  const ledger = res.body && res.body.ledger
-  if (res.statusCode !== 200 || !ledger) {
-    throw new ExternalError('Unable to identify ledger from account: ' + account)
-  }
-  return ledger
+function addressToLedger (address) {
+  return address.split('.').slice(0, -1).join('.')
```
clark800: Why do we want to have the notation for an account to be the same as the notation for a ledger? Isn't it more clear to indicate in the notation that the last component is an account? Then you would know that `a.b.c.d` is a ledger address and `a.b.c/d` is an account address.

bensharaf: The distinction between an account and a ledger only makes sense on an individual ledger, like the account `d` on the ledger `a.b.c`. In an inter-ledger context there's no real difference. Both can be routed to as hops to get to the final destination.

schemas/Quote.json
```diff
@@ -6,15 +6,15 @@
...
     "source_connector_account": {
       "description": "the connector's account on the source ledger",
-      "$ref": "Iri.json"
+      "type": "string"
     },
     "source_ledger": {
-      "description": "The URI of the source ledger",
-      "$ref": "Iri.json"
+      "description": "The ILP Address of the source ledger",
+      "type": "string"
```
emschwartz: I think this should reference a proper schema (https://github.com/interledger/five-bells-shared/pull/150/files#r71853250)

src/lib/route-builder.js
```diff
@@ -170,10 +174,13 @@ class RouteBuilder {
...
    * Round the hop's amounts according to the corresponding ledgers' scales/precisions.
    */
-  * _roundHop (hop, sourceUpDown, destinationUpDown) {
+  * _roundHop (hop, sourceUpDown, destinationUpDown, destinationPrecisionAndScale) {
     hop.sourceAmount = yield this._roundAmount('source', sourceUpDown, hop.sourceLedger, hop.sourceAmount)
-    hop.finalAmount = yield this._roundAmount('destination', destinationUpDown, hop.finalLedger, hop.finalAmount)
-    hop.destinationAmount = yield this._roundAmount('destination', destinationUpDown, hop.destinationLedger, hop.destinationAmount)
+    hop.destinationAmount = yield this._roundAmount('destination', destinationUpDown, hop.destinationLedger, hop.destinationAmount, destinationPrecisionAndScale)
+    // Only round the final amount if we know the precision/scale.
+    if (hop.isFinal || destinationPrecisionAndScale) {
```
emschwartz: Based on the comment above, shouldn't this be `&&`?

src/models/quote.js
emschwartz: One thing it does make slightly complicated is reasoning about how far down this path you would expect there to be connectors with exchange rates other than 1:1

test/data/paymentSameExecutionCondition.json
```diff
@@ -2,29 +2,29 @@
   "id": "c9377529-d7df-4aa1-ae37-ad5148612003",
```
emschwartz: If the `payment` data format isn't being used anymore the test data should probably be broken out into `transfer`s

src/models/quote.js
justmoon: The use case is I'm `us.fed.wf.stefan`. I want to make a subledger, but I don't want to worry about getting permission, peering with connectors, broadcasting routes etc. So I just call my subledger `us.fed.wf.stefan.` and assign addresses like `us.fed.wf.stefan.phone`. Every payment to `us.fed.wf.stefan.phone` is going to at some point show up in my account `us.fed.wf.stefan` unless somebody has a more direct route to my subledger. So I can run a connector against my account and route payments to my subledger successfully.

Subledgering is useful and extremely common. Banks run subledgers to the Fed, wallets usually run a subledger of whatever their bank is, etc. So it makes sense to have a format that makes subledgering natural and easy.

We could make a convention that we leave a trailing dot when we're talking about a ledger, i.e. `us.fed.wf.stefan` is referring to my account and `us.fed.wf.stefan.` is talking about my subledger.

src/controllers/quote.js
```diff
@@ -107,3 +108,9 @@ function validateAmounts (sourceAmount, destinationAmount) {
     }
   }
 }
+
+function validatePrecisionAndScale (precision, scale) {
+  if (precision && scale) return
+  if (!precision && !scale) return
```
justmoon: `scale = 0` is a valid value

src/models/quote.js
clark800: Thanks for the explanation. I'm still a bit confused though. If someone sent a payment to `us.fed.wf.stefan.phone` and the money shows up in the `us.fed.wf.stefan` account on the `us.fed.wf` ledger, what happens next? Does the money stay that account and the `stefan` subledger just keeps track of which sub-account the money belongs to?

src/controllers/quote.js
sentientwaffle: `precision` and `scale` here are from the query arguments, so they are `Strings`.

src/models/quote.js
justmoon: > If someone sent a payment to us.fed.wf.stefan.phone and the money shows up in the us.fed.wf.stefan account on the us.fed.wf ledger, what happens next?

I can run a connector which is watching my `us.fed.wf.stefan` account and will forward the payment onto my subledger.

clark800: I'm trying to understand what "forward the payment onto my subledger" means in terms of the actual balances on the ledgers. For a typical transfer the balance at `us.fed.wf.stefan` would go down and the balance at `us.fed.wf.stefan.phone` would go up the corresponding amount due to there being a third party market maker offsetting these changes. But in this case I imagine there is no third party market maker right? So if the `us.fed.wf.stefan` account balance goes down, how does that balance out on the `us.fed.wf` ledger?

justmoon: > But in this case I imagine there is no third party market maker right?

That's correct.

>  So if the us.fed.wf.stefan account balance goes down, how does that balance out on the us.fed.wf ledger?

The balance for `us.fed.wf.stefan` wouldn't go down. It would go *up*.

Somebody will prepare a transfer to me (as `us.fed.wf.stefan`) on the `us.fed.wf.` ledger. I (as the connector to my subledger) prepare a transfer out of the transactional account on the `us.fed.wf.stefan.` ledger to `us.fed.wf.stefan.phone`.

In plain English - if you send money to my phone, my bank account balance will be higher afterwards. If my phone spends money, my bank account balance will be lower.

clark800: Ok, so if I'm understanding correctly, the actual money stops flowing once it hits the `us.fed.wf.stefan` account, so for example a transfer from `us.fed.wf.stefan.phone` to `us.fed.wf.stefan.laptop` would not have any effect on the balance at `us.fed.wf.stefan` and thus isn't transferring any real money in the way that a transfer from `us.fed.wf.stefan` to `us.fed.boa.stefan` does (which requires counterparties). It seems like a significant semantic difference that the address notation abstracts. Does this abstraction provide benefits or does it potentially obfuscate the actual semantics? Comparing with gmail addresses, `account+alias@domain` (comparable to `stefan+phone@wf.fed.us`) there is a syntactic indicator for each semantic category. Since this is a fairly big decision, I think it might be worthwhile to have a design rationale to justify why our solution is preferable to the more conventional approach. Also, there are still outstanding questions on the original issue which may affect this pull request: https://github.com/interledger/rfcs/issues/31

src/lib/route-builder.js
sentientwaffle: No, `hop.isFinal` implies that the destination ledger is adjacent the the connector, so the connector can use the appropriate ledger plugin the discover the precision & scale.

`destinationPrecisionAndScale` only needs to be provided when the connector _cannot_ discover the precision & scale.

justmoon:
LGTM once tests pass

justmoon:
Before bumping the version, please let me rename `five-bells-connector` to `ilp-connector`. It makes sense to have the last version that uses URIs be called `five-bells-connector` and the first version that uses actual ILP addresses be called `ilp-connector`.

yana-novikova:
@justmoon can you address all our outstanding questions on this PR before merging it? This is fairly big protocol change and we want to have clear understanding of the reasoning behind these changes

src/models/quote.js
justmoon: > Ok, so if I'm understanding correctly, the actual money stops flowing once it hits the `us.fed.wf.stefan` account

From a clearing perspective, there is no such thing as "actual money". My personal ledger is a subledger of my bank's ledger which is a subledger of the Fed's ledger. If I'm transferring within my own ledger, it doesn't affect higher order ledgers. If I'm transferring within the bank's ledger, it doesn't affect higher order ledgers. If there was a subledger of my subledger and a transfer happened within that, e.g. `us.fed.wf.stefan.phone.foo` to `us.fed.wf.stefan.phone.bar`, it wouldn't appear on any higher order ledgers.

From a regulatory perspective, there *is* such a distinction ("real money" vs "pure accounting") and since regulatory issues are handled on the application layer, you may find functionality to differentiate that there.

Perhaps it's useful to show the difference between the clearing/interledger layer and application layer by example.

ILP addresses are on the Interledger layer and are used similarly to IP addresses:

ILP: `us.fed.wf.stefan`
IP: `172.217.21.174`

Note that these are low-level addresses intended primarily for use by the routing protocol. The routing protocol doesn't care about the difference between real money and fake money and so you don't see that distinction here.

On the application layer, you have human-facing identifiers:

SPSP: `alice@red.ilpdemo.org`
SMTP: `stefan@ripple.com`

Here, it's a totally different story. These are user-facing identifiers. We may need to know identities of real people for compliance purposes. I actually talked to @vhpoet and we have a tentative plan to add something exactly like you suggested `alice+phone@red.ilpdemo.org`

I think the thing that trips everybody up is that our low-level identifiers (ILP addresses) look like they're supposed to be human-readable: `us.fed.wf.bob`. We could also use some more IP address looking thing: `0123ff:213db1:123eba`. But while that would reduce some misunderstandings short-term I think in the long run it's going to make engineers' lives easier to be able to look at an ILP address and get some information from that without having to whois it.

(In my example `172.x.x.x` belongs to ARIN, i.e. it's something in North America. `172.217.x.x` belongs to Google. So it *is* possible to decode this hierarchical identifier. But because it's numbers it's a bit harder to read than, say, `na.google.web.frontend` would be.)

justmoon:
@yana-novikova Just replied! Apologies - I didn't see the question in this thread because Github collapsed that diff.

src/models/quote.js
clark800: > Note that these are low-level addresses intended primarily for use by the routing protocol. The routing protocol doesn't care about the difference between real money and fake money and so you don't see that distinction here.

This is helping me understand better what this is for. So where does the low-level address come from originally? If at the application layer the user enters an SPSP address like `alice@red.ilpdemo.org`, how does the application generate `us.fed.wf.alice`?

geertweening: @justmoon bump

justmoon: > So where does the low-level address come from originally? If at the application layer the user enters an SPSP address like alice@red.ilpdemo.org, how does the application generate us.fed.wf.alice?

SPSP is described in [IL-RFC-0009](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).

You first query webfinger to resolve the email-like identifier:

https://red.ilpdemo.org/.well-known/webfinger?resource=acct%3Abob%40red.ilpdemo.org

That tells us the SPSP endpoint URI `https://red.ilpdemo.org/api/receivers/bob`.

https://red.ilpdemo.org/api/receivers/bob

Querying that tells us the ILP address. Right now that's a `ledger, account` tuple, e.g. `ledger: 'https://red.ilpdemo.org/ledger', account: 'https://red.ilpdemo.org/ledger/accounts/alice'`. If this pull request lands, that would change to just be an ILP address, e.g. `ilpdemo.red.alice`.

If you're asking how the SPSP server (=receiver) knows its own address - right now that's just part of the SPSP server configuration. In the future we will likely want to come up with some auto-configuration scheme but that's out of scope for this PR.

clark800: Got it, thanks

PR #200, opened 2016-08-01
fix: make quoting less pessimistic

sentientwaffle:
Tests will pass once https://github.com/interledger/five-bells-integration-test/pull/27 is merged.

This change makes quoting less pessimistic by using unshifted local routes for `findBestHopBy*`, (but still broadcasting the shifted routes) and always rounding in the connector's favor.

### DIFF
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -16,7 +16,6 @@ const DEFAULT_SLIPPAGE = 0.001 // 0.1%
 const DEFAULT_ROUTE_BROADCAST_INTERVAL = 30 * 1000 // milliseconds
 const DEFAULT_ROUTE_CLEANUP_INTERVAL = 1000 // milliseconds
 const DEFAULT_ROUTE_EXPIRY = 45 * 1000 // milliseconds
-const DEFAULT_ROUTE_SHIFT = true
 
 function isRunningTests () {
   return (
@@ -201,8 +200,6 @@ function getLocalConfig () {
     Number(Config.getEnv(envPrefix, 'ROUTE_CLEANUP_INTERVAL')) || DEFAULT_ROUTE_CLEANUP_INTERVAL
   const routeExpiry =
     Number(Config.getEnv(envPrefix, 'ROUTE_EXPIRY')) || DEFAULT_ROUTE_EXPIRY
-  const routeShift =
-    Config.castBool(Config.getEnv(envPrefix, 'ROUTE_SHIFT'), DEFAULT_ROUTE_SHIFT)
 
   // Credentials should be specified as a map of the form
   // {
@@ -243,7 +240,6 @@ function getLocalConfig () {
     routeBroadcastInterval,
     routeCleanupInterval,
     routeExpiry,
-    routeShift,
     logLevel
   }
 }
--- a/src/lib/route-builder.js
+++ b/src/lib/route-builder.js
@@ -61,7 +61,8 @@ class RouteBuilder {
       _nextHop.sourceAmount = amountWithSlippage.toString()
       info.slippage = (amount - amountWithSlippage).toString()
     }
-    // Round in our favor to ensure that the adjustments don't fall off.
+    // Round in favor of the connector (source amount up; destination amount down)
+    // to ensure it doesn't lose any money. The amount is quoted using the unshifted rate.
     const nextHop = yield this._roundHop(_nextHop, 'up', 'down',
       query.destinationPrecisionAndScale)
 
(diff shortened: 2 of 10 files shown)


src/lib/routing-tables.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  removeExpiredRoutes () {
+    this.localTables.removeExpiredRoutes()
+    this.publicTables.removeExpiredRoutes()
+  }
+
+  * _getScaleAdjustment (infoCache, ledger) {
+    const scale = (yield infoCache.get(ledger)).scale
+    return scale ? (1 / Math.pow(10, scale)) : 0
```
clark800: Might be more efficient to do `Math.pow(10, -scale)`

sentientwaffle: I'm not too concerned about the efficiency, but the negative exponent is more straightforward -- I'll change it.

src/lib/config.js
```diff
@@ -201,8 +200,6 @@ function getLocalConfig () {
     Number(Config.getEnv(envPrefix, 'ROUTE_CLEANUP_INTERVAL')) || DEFAULT_ROUTE_CLEANUP_INTERVAL
   const routeExpiry =
     Number(Config.getEnv(envPrefix, 'ROUTE_EXPIRY')) || DEFAULT_ROUTE_EXPIRY
-  const routeShift =
```
emschwartz: Taking out this option probably counts as a breaking change

src/lib/route-builder.js
```diff
@@ -110,8 +110,8 @@ class RouteBuilder {
     const _nextHopBySourceAmount = this.routingTables.findBestHopForSourceAmount(
       sourceLedger, finalLedger, sourceTransfer.amount)
     if (!_nextHopBySourceAmount) throwAssetsNotTradedError()
-    // Round against ourselves since the quote rate was overestimated in our favor.
-    const nextHop = yield this._roundHop(_nextHopBySourceAmount, 'down', 'up')
+    // Round in favor of the connector.
+    const nextHop = yield this._roundHop(_nextHopBySourceAmount, 'up', 'down')
```
emschwartz: Could you add some more comments to explain this more? Also, why is this changing from `'down', 'up'` to `'up', 'down'`? It would be to clarify more the thinking behind the logic for the next person who has to modify this code

src/lib/routing-tables.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+const routing = require('five-bells-routing')
+
+/**
+ * Maintain two sets of `routing.RoutingTables`.
```
emschwartz: Is there a way to modify the code in `five-bells-routing` to make the raw vs shifted option configurable? Seems like that would be better than copying over this code here

src/lib/config.js
sentientwaffle: The changes should make the `routeShift` option unnecessary, so removing it doesn't actually break anything.

src/lib/route-builder.js
sentientwaffle: I'll add more comments. The short answer as to why it changed: before it rounded in favor of the sender/receiver because the route shifting had already occurred, so despite the rounding the result was still overall in favor of the connector. Now, though, the local part of the route hasn't been shifted, so in order to avoid losing money the connector must round in its own favor.

src/lib/routing-tables.js
sentientwaffle: Despite having the same name, the `RoutingTables` is a bit different from the one in five-bells-routing.

src/lib/routing-tables.js
```diff
@@ -0,0 +1,70 @@
...
+ * The "local" tables consist of the raw local routes, joined to the
+ * remote (shifted) routes.
+ *
+ * The "public" tables consist of the local shifted routes, joined to the
+ * remote (shifted) routes.
+ * Before joining, the local routes are shifted down by 1/10^destination_ledger_scale
+ * to pessimistically account for rounding errors.
+ *
+ * The "local" tables are used for `findBestHopFor*`.
+ * The "public" tables are broadcast to adjacent connectors.
```
emschwartz: Could you add a bit more explanation here about why there are two sets?

emschwartz:
LGTM

src/lib/routing-tables.js
emschwartz: How about:


> When routing payments across multiple ledgers, each hop will round the amounts according to the precision of the ledger. These rounding errors can accumulate to the point where a connector later in a path may receive an amount that is less than the rate they are willing to accept.
> 
> Connectors broadcast slightly pessimistic rate curves in order to compensate for rounding errors across multi-hop routes. They shift the curve by the maximum amount that could be lost due to rounding errors, so that even if the incoming amount is rounded down by that amount the resulting value will still match the connector's minimum rate.
> 
> Connectors use unshifted rate curves locally to determine whether an incoming payment request matches their minimum rate. When applying this rate, they round in their favor to insure that they never accept a payment that is lower than their rate.
> 
> This class maintains two sets of `routing.RoutingTables`: the "local" tables are the optimistic ones and the "public" tables are the pessimistic ones.
> 
> The "local" tables consist of the raw local routes, joined to adjacent connector's (public, shifted) routes. They are used for `findBestHopFor*`, the results of which must be rounded in the connector's own favor to ensure that they don't lose money.
> 
> The "public" tables consist of the local shifted routes, joined to the remote (shifted) routes. Before joining, the local routes are shifted down by 1/10^destination_ledger_scale to account for rounding errors. Their curves are broadcast to adjacent connectors.

Issue #198, opened 2016-07-26
Move `ledger` property assignment to plugins and remove `getPrefix`


justmoon:
Remove this code:

https://github.com/interledger/five-bells-connector/blob/f223d7df1f2cd3820ad318832d5dff3bc6189eb1/src/models/subscriptions.js#L12

https://github.com/interledger/five-bells-connector/blob/f223d7df1f2cd3820ad318832d5dff3bc6189eb1/src/models/subscriptions.js#L21

and make it the responsibility of the plugins to assign the `ledger` property.

justmoon:
Note this change affects at least `ilp-plugin-tests`, `ilp-plugin-bells`, `ilp-plugin-virtual` and `ilp-connector`.

sentientwaffle:
Resolved in https://github.com/interledger/five-bells-connector/pull/202

PR #202, opened 2016-08-02
fix: assume transfer.ledger property is assigned by plugins

sentientwaffle:
* Depends on https://github.com/interledger/js-ilp-plugin-bells/pull/20
* Part of https://github.com/interledger/five-bells-connector/issues/198


emschwartz:
LGTM

PR #203, opened 2016-08-04
fix: [BREAKING] make plugin options flat

bensharaf:
Addresses issue interledger/rfcs#55 . Instead of separating pluginOptions into auth and other options, the structure is flattened and special fields (such as store) are prefixed with an underscore. This depends on https://github.com/interledger/js-ilp-core/pull/31 , https://github.com/interledger/js-ilp-plugin-bells/pull/22 , and https://github.com/interledger/js-ilp-plugin-virtual/pull/5


bensharaf:
The checks are going to fail until the pull requests to `ilp-plugin-bells` and `ilp-core` regarding this issue are merged.

src/lib/core.js
```diff
@@ -15,14 +15,14 @@ module.exports = function (options) {
     const store = creds.store && newSqliteStore(creds.store)
 
     creds.prefix = ledgerPrefix
-    core.addClient(ledgerPrefix, new ilpCore.Client({
-      plugin: require('ilp-plugin-' + creds.type),
-      auth: creds,
-      store: store,
-      log: makeLogger('plugin-' + creds.type),
+    core.addClient(ledgerPrefix, new ilpCore.Client(Object.assign({
```
justmoon: It seems like giving the built-in options a higher precendence would be wise. `Object.assign({}, creds, {`

emschwartz:
LGTM

PR #206, opened 2016-08-16
feat: relay quotes for unknown ledgers

sentientwaffle:
Depends on https://github.com/interledger/js-ilp-core/pull/39 and https://github.com/interledger/five-bells-routing/pull/8


justmoon:
LGTM. Lots of good refactoring going on there. :muscle:

test/quoteSpec.js
```diff
@@ -668,15 +668,15 @@ describe('Quotes', function () {
...
+    it('fails on a same-ledger quote', function * () {
       yield this.request()
         .get('/quote?source_amount=100' +
           '&source_address=usd-ledger.alice' +
           '&destination_address=usd-ledger.bob')
         .expect(422)
         .expect(function (res) {
           expect(res.body.id).to.equal('AssetsNotTradedError')
-          expect(res.body.message).to.match(/source_ledger must be different from destination_ledger/)
+          expect(res.body.message).to.match(/This connector does not support the given asset pair/)
```
emschwartz: It might be nice to throw a more descriptive error here

michielbdejong:
The title of this PR was 'relay quotes for unknown ledgers', but looking at [the changeset](https://github.com/interledgerjs/ilp-connector/pull/206/files#diff-d1efbef0a5a24c11beea040a0448ad2dL51), it seems to me that what it does is to relay quotes for *all* ledgers, not just unknown ones? Shouldn't it first look in its own routing tables, and then only do a remote quote in case nothing is found, or in case a (default) route is found but it has no curve?

Issue #205, opened 2016-08-15
Set Interledger network up


kopnin:
@emschwartz 
Hi Evan, 
I would like to connect 2 private ledgers. Interledger seems very suitable for that. I've run five-bells-demo and looked into the code. However,  I still have misunderstanding. You use *js-ilp-plugin-bells* and *five-bells-connector*. Unfortunately, I didn't find unified connector implementation (only for five-bells demo). To achieve my plans I need my own plugin and unified connector as well, isn't it? 

Regarding plugin: I know you already have some different plugins, cool. What kind of plugins do you plan to implement and when?

emschwartz:
The `five-bells-connector` will soon be renamed `js-ilp-connector` because it is meant to be used with different ledgers. In order to use the connector for a different type of ledger you would need to implement one of those plugins for your ledger but you shouldn't need to reimplement the connector.

Using the connector for non-five-bells-ledgers hasn't been tested too much yet so it's likely there are bugs waiting to be found. I know @diminator has been working on this.

I'm hoping we can implement plugins for the major cryptocurrencies, but we're a bit low on bandwidth right now so no promises about when that will be done. If you know anyone that would be interested in working on this, please tell them to get in touch with me.

kopnin:
Thanks, 
Yes, I know that I haven't to reimplement the connect, only plugin. Evan, Is it means that I can reuse five-bells-connector right now? Or I should wait some time while you will rename it. I mean when you will rename it will you extend/change functionality of connector?

emschwartz:
You can start working with it now. The rename won't affect the functionality, only better reflect that it will no longer be limited to the `five-bells` projects (i.e. the `five-bells-ledger`)

kopnin:
Cool! Thank you, Evan

PR #207, opened 2016-08-16
chore: rename five-bells-connector to ilp-connector

justmoon:
Align this project with our updated naming scheme. We decided a few months ago that "Five Bells" was going to be the name of Five Bells Ledger, but all generic parts of the reference implementation should be called just `ilp-*`, e.g. `ilp-core`. Additionally, we prefixed all JavaScript repo names (but not npm packages) with `js-`.

So the repository name for this repository will be `js-ilp-connector` whereas the NPM package will be `ilp-connector`.

The rename is intended to alleviate confusion as in #205


emschwartz:
LGTM once integration tests are updated and the build is passing

PR #208, opened 2016-08-17
fix: Plugin#getAccount returns a Promise

sentientwaffle:
Requires https://github.com/interledger/js-ilp-core/pull/40 and https://github.com/interledger/js-ilp-plugin-bells/pull/27


emschwartz:
LGTM once tests pass

PR #209, opened 2016-08-26
feat: add a backend using coinmarketcap

justmoon:
Load exchange rate data from Coinmarketcap (and combine with Fixer.io) to support cryptocurrencies and fiat currencies alike.


bensharaf:
LGTM

PR #210, opened 2016-08-26
fix: remove credential validation

bensharaf:
The fields that `validateCredentialsEnv` checked were specific to `ilp-plugin-bells`, and shouldn't be required for other plugin types. Validating credentials should be the responsibility of the ledger plugins.


emschwartz:
Please add a test or two to make sure it handles other credentials correctly

justmoon:
LGTM

PR #211, opened 2016-08-31
feat: relay rejection from destination to source

sentientwaffle:
Depends on https://github.com/interledger/js-ilp-plugin-bells/pull/34.


test/paymentsSpec.js
```diff
@@ -253,6 +253,21 @@ describe('Payments', function () {
     }
   })
 
+  it('relays a rejection', function * () {
```
emschwartz: Shouldn't it relay the actual message the destination transfer was rejected with?

emschwartz:
LGTM

PR #214, opened 2016-09-08
feat: BREAKING don't autoload peer connectors by default

### LINKED ISSUE #212: Add ability to add connectors through configuration
The `getConnectors` method on the ledger plugin should be viewed as an automatic configuration (similar to `getPrefix`) and connectors should be able to override it via their configuration.

We should assume that most connectors will pass in a list of connectors via the config and using `getConnectors` should be the exception (although still supported.)

Proposed config

``` sh
# Provide a basic list of peers
CONNECTOR_PEERS=https://connie.ilpdemo.org,https://fred.example.com # default = ''

# Whether to automatically populate the list of peers by calling getConnectors
# on all ledger plugins
CONNECTOR_AUTOLOAD_PEERS=true # default = false
```

sentientwaffle:
Fixes #212

Depends on https://github.com/interledger/js-ilp-core/issues/48


emschwartz:
LGTM once tests pass

PR #217, opened 2016-09-12
Fix Dockerfile.

earizon:
Install npm dependencies in docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@ WORKDIR /usr/src/app
 
 COPY . /usr/src/app
 
+RUN cd /usr/src/app && npm install
+
 EXPOSE 3000
 
 CMD [ "npm", "start" ]


justmoon:
Hmm. On the ledger, we actually intentionally don't do `npm install` in the Dockerfile. Instead we do `npm install` before running `docker build`, so that the `node_modules

earizon:
Got it!. Guess the best way to do it would be creating an specific docker volume container with the node js external modules as explained in "Creating and mounting a data volume container" at https://docs.docker.com/engine/tutorials/dockervolumes/. 

 This allows Docker composer to make replicable installations and this same strategy can also be useful with other reusable volume instances (shared volumes with test-data for example). Before re-inventing the wheel, is there something like this already in place?

justmoon:
I think you'd just put it in the Docker image, not a volume. (The npm deps are part of the code + dependencies, which is what should be immutable and part of the image.)

Basically, what the Dockerfile does now is correct imo - you just need to run `npm install` before you run `docker build`. I'll add something to the README that explains that.

PR #218, opened 2016-09-13
fix: don't shift the Route for the trivial rate

### LINKED ISSUE #216: Don't round down when not necessary
Here's the feedback I got

> I've set CONNECTOR_FX_SPREAD=0 CONNECTOR_SLIPPAGE=0.

> When I transfer £10 from wallet a to wallet b,  £10 is sent from a but only £9.99 is received by wallet b. Is there another parameter to change that could account for the £0.01 lost or is this a rounding error?

sentientwaffle:
Fixes #216


justmoon:
LGTM

PR #221, opened 2016-09-20
chore: publish docker image

emschwartz:
`DOCKER_{EMAIL,USER,PASS}` env vars also need to be added to circle for this to work


circle.yml
```diff
@@ -19,9 +21,9 @@ deployment:
       - git config --global user.name "CircleCI"
       - git config --global push.default simple
       - node scripts/publish_web.js
-      - mv npmrc-env .npmrc
```
justmoon: Why not just keep this? Then you don't need to have the `.npmrc` in the repo normally.

emschwartz: 👍

PR #222, opened 2016-09-20
chore: update ilp-plugin-virtual

bensharaf:
Update plugin virtual.


bensharaf:
The integration tests fail, but master is failing integration tests already. `ilp-plugin-virtual` isn't used in the integration tests.

PR #225, opened 2016-09-26
Fix: update node version in docker to 6

bensharaf:
Updates the node version used in the `Dockerfile`.


bensharaf:
@emschwartz: Addressed your changes.

PR #233, opened 2016-10-03
Feat: [BREAKING] improve PAIRS, LEDGERS, CREDENTIALS

bensharaf:
Removes `CONNECTOR_CREDENTIALS`, instead rolling those options into `CONNECTOR_LEDGERS`. Fixes a problem in the auto-generation of `CONNECTOR_PAIRS` which causes pairs to be generated in one direction only (would include `usd -> eur` but not `eur -> usd`). Updates the README to reflect these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,31 +47,32 @@ npm start
 
 #### Trading
 
-* `CONNECTOR_LEDGERS` (default: `[]`) Ledgers where this connector has accounts. Used to auto-generate `CONNECTOR_PAIRS`.
-```js
-[
-  "USD@example.usd-ledger.",
-  "EUR@example.eur-ledger."
-]
-```
-* `CONNECTOR_CREDENTIALS` (default: `{}`) Connector's login credentials for various ledgers, ex.
+* `CONNECTOR_LEDGERS` (default: `{}`) Connector's login credentials for ledgers where it has accounts. Used to auto-generate `CONNECTOR_PAIRS`.
 ```js
 {
   // Using Basic Auth
-  "<ledger_address>": {
-    "account": "...",
-    "username": "...",
-    "password": "..."
-    "ca": "...", // Optional
+  "example.usd-ledger.": {
+    "currency": "USD", // asset on this ledger
+    "plugin": "ilp-plugin-bells", // module for this ledger plugin
+    "options": { // actual plugin options passed into plugin constructor
+      "account": "...",
+      "username": "...",
+      "password": "..."
+      "ca": "...", // Optional
+    }
   },
 
   // Using Client Certificate Auth
-  "<ledger_address_2>": {
-    "account": "...",
-    "username": "...",
-    "cert": "...",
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -47,26 +47,19 @@ npm start
...
-  "EUR@example.eur-ledger."
-]
-```
-* `CONNECTOR_CREDENTIALS` (default: `{}`) Connector's login credentials for various ledgers, ex.
+* `CONNECTOR_LEDGERS` (default: `{}`) Connector's login credentials for ledgers where it has accounts. Used to auto-generate `CONNECTOR_PAIRS`.
 ```js
 {
   // Using Basic Auth
-  "<ledger_address>": {
+  "USD@example.usd-ledger.": {
```
emschwartz: We should get rid of the `USD@` part and just put a key in the object for `currency`

README.md
```diff
@@ -47,26 +47,19 @@ npm start
...
 
-* `CONNECTOR_LEDGERS` (default: `[]`) Ledgers where this connector has accounts. Used to auto-generate `CONNECTOR_PAIRS`.
-```js
-[
-  "USD@example.usd-ledger.",
-  "EUR@example.eur-ledger."
-]
-```
-* `CONNECTOR_CREDENTIALS` (default: `{}`) Connector's login credentials for various ledgers, ex.
+* `CONNECTOR_LEDGERS` (default: `{}`) Connector's login credentials for ledgers where it has accounts. Used to auto-generate `CONNECTOR_PAIRS`.
```
emschwartz: This should be a breaking feature instead of a fix

src/lib/utils.js
```diff
@@ -3,18 +3,19 @@
 const crypto = require('crypto')
 
 /**
- * Get all possible pair combinations from an array.
+ * Get all possible pair combinations from an array, order sensitive.
```
emschwartz: If I don't want to support all possible pairs, how would I express that?

bensharaf: you can still manually specify pairs. This just makes the default behavior actually do what it claims.

README.md
bensharaf: I definitely agree that the `asset@ledger` format is weird. Adding `currency` (or `_currency` to the object could mean that it gets passed into the plugin options even though the plugin shouldn't care what the connector calls its asset.

README.md
bensharaf: renamed the PR

README.md
emschwartz: Oh, it should probably be made clearer that those maps are going to be passed directly to the plugins. Another way to handle this would be:

```js
"ledger.address.": {
  "currency": "USD",
  "plugin": "ilp-plugin-bells",
  "options": {
    // options passed directly to the plugin
  }
}
```
An advantage of this would be it would be easier to add ledger-specific configuration that is used by the connector but not the plugin.

PR #235, opened 2016-10-07
Feat: modify plugins

bensharaf:
Support the adding and removing of plugins while a connector is running. Two new functions are exposed in `index.js`:

- `addPlugin(id, options, tradesTo, tradesFrom)` - connect a new plugin on ledger `id` with options `options`. The options are in the same format as the values in the `CONNECTOR_LEDGERS` object. `tradesTo` and `tradesFrom` are optional should be a list in the form `['USD@other.ledger.', 'EUR@another.ledger.', ..]`. If they aren't specified, the new ledger is paired with every other ledger in the pairs.

- `removePlugin(id)` disconnects ledger `id` and removes it from the routing tables and trading pairs.

Depends on https://github.com/interledger/js-ilp-routing/pull/10


src/app.js
```diff
@@ -71,7 +73,39 @@ function listen (koaApp, config, core, backend, routeBuilder, routeBroadcaster)
...
+      connector: config.server.base_uri,
+      _plugin: require(options.plugin),
+      _log: log.create(options.plugin)
+    })))
+
+    yield core.getClient(id).connect()
+
+    if (tradesTo) {
+      tradingPairs.addPairs(tradesTo.map((e) => [options.currency + '@' + id, e]))
+    } if (tradesFrom) {
```
justmoon: I would put the new `if` on a separate line to make it clearer that they're separate. Right now it's easy to confuse this with an `else if` due to the unusual formatting.

src/lib/trading-pairs.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
+const _ = require('lodash')
+
+class TradingPairs {
+  constructor (pairs) {
+    this._pairs = pairs
+  }
+
+  getPairs () {
```
justmoon: Maybe this should be `toArray`. As is, you have things like `pairs.getPairs` which is confusing: "I thought I already had pairs, why would call `getPairs` on `pairs`?"

PR #239, opened 2016-10-21
Feat: add custom routes in CONNECTOR_ROUTES

sharafian:


src/lib/config.js
```diff
@@ -116,13 +116,27 @@ function parseLedgers () {
   })
 }
 
+function parseRoutes () {
+  return JSON.parse(Config.getEnv(envPrefix, 'ROUTES') || '[]')
```
emschwartz: Might be good to validate the data that's passed in, just so you throw an error immediately if someone gives invalid configuration rather than waiting until it causes a problem when the connector tries to use it

src/lib/route-broadcaster.js
```diff
@@ -107,6 +109,31 @@ class RouteBroadcaster {
...
   }
 
+  addConfigRoutes () {
+    let that = this
+
+    return co(function * () {
+      for (let configRoute of that.configRoutes) {
+        const connectorLedger = configRoute.connector
+          .split('.')
+          .slice(0, configRoute.connector.split('.').length - 1)
```
sentientwaffle: This can just be `.slice(0, -1)`.

src/lib/route-broadcaster.js
```diff
@@ -107,6 +109,31 @@ class RouteBroadcaster {
...
+          .join('.') + '.'
+
+        const route = new Route(
+          // use a 1:1 curve as a placeholder (it will be overwritten by a remote quote)
+          [ [0, 0], [1, 1] ],
+          // the ledger '.' is inserted to make sure that the hop to the connectorLedger
+          // is not considered final.
+          [ connectorLedger, '.' ],
```
sentientwaffle: This will cause issues when broadcasting routes. The broadcasted route will have a `destinationLedger` of `"."`, which doesn't make sense. Should it be `configRoute.prefix` or similar instead?

README.md
```diff
@@ -89,6 +89,20 @@ npm start
 ]
 ```
 
+* `CONNECTOR_ROUTES` (default: `[]`) Explicitly add routes to the connector. If `prefix` is the most specific
+  route name that matches a destination, then the payment will be forwarded to `connector`. ex:
+```js
+[
+  {
+    "prefix": "", // matches any destination
+    "connector": "ilpdemo.red.connie"
```
sentientwaffle: I don't think the connector should assume that the ledger prefix of `"ilpdemo.red.connie"` is `"ilpdemo.red."`. I'd suggest renaming the `connector` field to `connectorAccount`, and adding a `connectorLedger` field as well.

sentientwaffle: And, for clarity, rename the `prefix` field (e.g. `destinationPrefix` or `targetPrefix`).

sentientwaffle: The latter.

sharafian: changed to the new field names

src/lib/route-broadcaster.js
sharafian: I think that of the possible options, `destinationPrefix` makes the most sense. changed.

src/lib/config.js
sharafian: added some validation

README.md
```diff
@@ -89,6 +89,22 @@ npm start
...
 ```
 
+* `CONNECTOR_ROUTES` (default: `[]`) Explicitly add routes to the connector. If `prefix` is the most specific
+  route name that matches a destination, then the payment will be forwarded to `connector`. ex:
+```js
+[
+  {
+    "destinationPrefix": "", // matches any destination
+    "connectorAccount": "ilpdemo.red."
+    "connectorLedger": "ilpdemo.red.connie"
```
sentientwaffle: `connectorAccount` and `connectorLedger` are swapped

src/lib/config.js
```diff
@@ -116,13 +116,44 @@ function parseLedgers () {
   })
 }
 
+function parseRoutes () {
+  const routes = JSON.parse(Config.getEnv(envPrefix, 'ROUTES') || '[]')
+
+  for (let route of routes) {
+    if (
+      !route.connectorLedger.match(/^[a-zA-Z0-9._~-]+\.$/) ||
+      typeof route.connectorLedger !== 'string' ||
```
sentientwaffle: You should check if its a string _before_ matching it against a regexp, not after.

src/lib/routing-tables.js
```diff
@@ -62,9 +62,9 @@ class RoutingTables {
     }
   }
 
-  addRoute (route) {
-    this.localTables.addRoute(route)
-    return this.publicTables.addRoute(route)
+  addRoute (route, prefix) {
```
sentientwaffle: `destinationPrefix`, to be more precise.

test/routeBroadcasterSpec.js
```diff
@@ -81,6 +90,19 @@ describe('RouteBroadcaster', function () {
     })
   })
 
+  describe('addConfigRoutes', function () {
+    it('loads routes from CONNECTOR_ROUTES', function * () {
+      this.broadcaster.addConfigRoutes()
```
sentientwaffle: Doesn't this need to be `yield`ed so that it actually runs?

src/lib/config.js
sharafian: fixed.

README.md
```diff
@@ -89,6 +89,22 @@ npm start
 ]
 ```
 
+* `CONNECTOR_ROUTES` (default: `[]`) Explicitly add routes to the connector. If `prefix` is the most specific
```
sentientwaffle: `targetPrefix`

README.md
```diff
@@ -89,6 +89,22 @@ npm start
...
+  route name that matches a destination, then the payment will be forwarded to `connector`. ex:
```
sentientwaffle: `connectorAccount`

PR #240, opened 2016-10-24
Switch logging from bunyan to riverpig

justmoon:
Bunyan is causing issues during development due to the separate CLI transformer process. Riverpig is a cleaner, simpler logger. As a dependency it is 10x smaller than bunyan.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,9 +48,10 @@
     "moment": "^2.10.2",
     "node-uuid": "^1.4.2",
     "reconnect-core": "^1.2.0",
-    "riverpig": "^1.0.2",
+    "riverpig": "^1.1.0",
     "spdy": "^3.2.3",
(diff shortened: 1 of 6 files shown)


justmoon:
@emschwartz I added a readme to riverpig.

emschwartz:
@justmoon thanks. I'm still not totally clear why riverpig is better than <insert other popular logging module here>

justmoon:
I looked a bunch of logging modules including `bunyan`, `winston`, `mag` (which we were using before) and `debug-levels`. They all either didn't integrate with `debug` well, weren't lightweight (Bunyan + dependencies for example is 10x the size of Riverpig + dependencies) or didn't fit within twelve-factor principles.

Note that logging in twelve-factor apps is a very small function, so I'm not so concerned about having to maintain our own module.

PR #242, opened 2016-10-27
fix: remove http; forward routes filtered by ledger

sentientwaffle:
Depends on https://github.com/interledgerjs/five-bells-integration-test/pull/52


src/lib/route-broadcaster.js
```diff
@@ -65,14 +65,21 @@ class RouteBroadcaster {
...
+        routes.filter((route) => route.source_ledger === adjacentLedger))
+    }))
+  }
+
+  _broadcastToLedger (adjacentLedger, routes) {
+    const connectors = Object.keys(this.peersByLedger[adjacentLedger])
+    return Promise.all(connectors.map((adjacentConnector) => {
+      const account = adjacentLedger + adjacentConnector
+      log.info('broadcasting ' + routes.length + ' routes to ' + account)
+      return this.core.getPlugin(adjacentLedger).sendMessage({
```
sharafian: Is there a protocol defined for these inter-connector messages?

README.md
```diff
@@ -40,9 +40,6 @@ npm start
 
 #### General
 
-* `CONNECTOR_BIND_IP` (default: `0.0.0.0`) IP that ILP Connector will bind to.
```
sharafian: The `CONNECTOR_PEERS` field should also be updated for these changes

src/app.js
```diff
@@ -2,20 +2,8 @@
 
 const _ = require('lodash')
 const co = require('co')
-const metadata = require('./controllers/metadata')
```
sharafian: Is there new way to get metadata/health/pairs with the connector messaging? It seems like you should still be able to query health and trading pairs

sentientwaffle: Not at the moment. Pairs is only used by the visualization, and I removed it in https://github.com/interledgerjs/five-bells-visualization/pull/35. Metadata isn't needed anymore since there are no URLs.

@justmoon do you think any of these features should be replicated with `sendMessage`/`incoming_message`, or are they no longer needed?

src/lib/route-broadcaster.js
sentientwaffle: For quoting there is https://github.com/interledger/rfcs/blob/master/0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md, but there's no spec for route broadcasting yet.

PR #246, opened 2016-11-10
feat: Added a warn message if CONNECTOR_MIN_MESSAGE_WINDOW is set to …

dappelt:
…a very small value.

Setting the env. variable CONNECTOR_MIN_MESSAGE_WINDOW to a very small value is likely to cause the

loss of funds, because the connector has not enough time to pass on the fulfillment.

### DIFF
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -4,6 +4,8 @@ const fs = require('fs')
 const Config = require('five-bells-shared').Config
 const Utils = require('../lib/utils')
 const _ = require('lodash')
+const logger = require('../common/log')
+const log = logger.create('config')
 
 const envPrefix = 'CONNECTOR'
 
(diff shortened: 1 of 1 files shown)


src/lib/config.js
```diff
@@ -167,6 +169,11 @@ function getLocalConfig () {
   const expiry = {}
   expiry.minMessageWindow =
     +Config.getEnv(envPrefix, 'MIN_MESSAGE_WINDOW') || DEFAULT_MIN_MESSAGE_WINDOW
+
+  if (expiry.minMessageWindow < 1) {
+    log.warn('CONNECTOR_MIN_MESSAGE_WINDOW is less than the recommended value. Set it to a greater value to avoid lossing funds.')
```
emschwartz: It would be good to explain what the problem with a low value here would be, or point to somewhere with more of an explanation. I might say something like `CONNECTOR_MIN_MESSAGE_WINDOW is less than the recommended value of 1 second. Short message windows increase the likelihood that the connector will be unable to fulfill incoming transfers before their expiries, resulting in a loss of funds. For more information on this risk, see the Interledger whitepaper`

PR #251, opened 2016-11-18
fix: allow BACKEND to be an external module name

sentientwaffle:
Fixes https://github.com/interledgerjs/ilp-connector/issues/249

This doesn't address the issue that Backends should supply route curves.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -165,14 +165,17 @@ function createApp (config, core, backend, routeBuilder, routeBroadcaster, routi
 }
 
 function getBackend (backend) {
+  if (moduleExists('./backends/' + backend)) return require('./backends/' + backend)
+  if (moduleExists(backend)) return require(backend)
+  throw new Error('Backend not found at "' + backend + '" or "/backends/' + backend + '"')
+}
+
+function moduleExists (path) {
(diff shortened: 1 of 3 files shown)


src/app.js
```diff
@@ -169,4 +164,16 @@ function createApp (config, core, backend, routeBuilder, routeBroadcaster, routi
   }
 }
 
+function getBackend (backend) {
+  try {
+    return require('./backends/' + backend)
+  } catch (err) {
```
justmoon: This will catch all errors, which can be very confusing. Suppose some module that your backend relies on throws when you require it. The error you'll get is "Backend not found", but you check and the backend is definitely, definitely there.

The solution is to first call `require.resolve` which will *only* fail if the immediate dependency couldn't be found. Then require the actual module outside of the try/catch so that any other errors will bubble up normally.

For example:

```js
function loadModule (modulePath) {
  try {
    require.resolve(modulePath)
  } catch (err) {
    throw new Error('Module "' + modulePath + '" not found')
  }

  return require(modulePath)
}
```

sentientwaffle: Fixed

PR #256, opened 2016-11-29
fix: disable plugin.connect() timeout

sentientwaffle:
Disable the timeout introduced by https://github.com/interledgerjs/ilp-plugin-bells/pull/74.

The timeout should be disabled because the ledger may just not have been started yet (see https://github.com/interledgerjs/ilp-connector/issues/248).

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -49,7 +49,7 @@ function addPlugin (config, core, backend, routeBroadcaster, tradingPairs, id, o
       _log: logger.createRaw(options.plugin)
     })))
 
-    yield core.getClient(id).connect()
+    yield core.getClient(id).connect({timeout: 0})
 
     if (tradesTo) {
       tradingPairs.addPairs(tradesTo.map((e) => [options.currency + '@' + id, e]))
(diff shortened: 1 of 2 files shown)


src/lib/core.js
```diff
@@ -17,6 +17,8 @@ module.exports = function (options) {
     creds.options.prefix = ledgerPrefix
     core.addClient(ledgerPrefix, new ilpCore.Client(Object.assign({}, creds.options, {
       debugReplyNotifications: config.features.debugReplyNotifications,
+      // disable connect() timeout
+      connectTimeout: 0,
```
emschwartz: This suggests that `connectTimeout` should be a standardized parameter in the ledger plugin interface. Is that right?

emschwartz: I think so. Also add it to the ledger plugin interface spec. If you're passing it to the ledger plugin no matter what type it is, and if we think it's a useful options for other plugins to have, it should be part of the standard options.

emschwartz:
Is there a way to make sure the connector does not serve rates/quotes for a specific ledger when it is not connected? I'm thinking both before it connects initially as well as if it detects that it gets disconnected later, though the latter may be more difficult.

sentientwaffle:
@emschwartz ok, updated.

> Is there a way to make sure the connector does not serve rates/quotes for a specific ledger when it is not connected? I'm thinking both before it connects initially as well as if it detects that it gets disconnected later, though the latter may be more difficult.

That's going to be addressed in a separate PR (it overlaps with https://github.com/interledgerjs/ilp-connector/issues/248).

PR #263, opened 2016-12-02
fix: fix Quote.json schema

sentientwaffle:
Fixes https://github.com/interledgerjs/ilp-connector/issues/262

### DIFF
--- a/src/models/quote.js
+++ b/src/models/quote.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const isUndefined = require('lodash/fp/isUndefined')
+const omitUndefined = require('lodash/fp/omitBy')(isUndefined)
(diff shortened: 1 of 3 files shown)


src/models/quote.js
```diff
@@ -65,7 +67,7 @@ function * getQuote (params, config, routeBuilder, balanceCache) {
     quote.destinationExpiryDuration,
     quote.minMessageWindow, config)
 
-  return {
+  return omitUndefined({
```
dappelt: Why is the call to omitUndefined(...) necessary? Is there any scenario where the argument to this function can be undefined?

sentientwaffle: `additional_info` can be `undefined`.

dappelt: Ok, makes sense.

PR #273, opened 2016-12-09
Feat: add better plugin store

sharafian:
`ilp-plugin-virtual` needs a store in order to be used by the connector. The current connector uses a sqlite database in a different file for each plugin. This is a little messy; the store should probably be able to go into an ILP Kit's database to make everything cleaner.

This new plugin store uses Sequelize in order to support a store that connects through either sqlite or postgres.

Unfortunately, the `sqlite3` module is still required, because it is needed for testing.

There is a new variable, `CONNECTOR_DB_URI`, which should be set to the same one as the API and Ledger by the ILP Kit.


src/lib/core.js
```diff
@@ -12,7 +12,15 @@ module.exports = function (options) {
...
     const creds = _.cloneDeep(config.ledgerCredentials[ledgerPrefix])
-    const store = creds.options.store && newSqliteStore(creds.options.store)
+    let store = null
+
+    if (creds.store) {
+      if (!config.databaseUri) {
+        throw new Error('missing CONNECTOR_DB_URI; cannot create plugin store for ' + ledgerPrefix)
+      }
+      store = new PluginStore(config.databaseUri, ledgerPrefix)
+      store.connect()
```
sentientwaffle: Isn't this async?

package.json
```diff
@@ -43,7 +43,7 @@
     "lodash": "^4.6.1",
     "moment": "^2.10.2",
     "riverpig": "^1.1.0",
-    "sqlite3": "^3.1.4",
+    "sequelize": "^3.27.0",
```
vhpoet: We depend on `sequelize` anyway, `ILP Kit` uses `sequelize`. I think the only reason for using `knex` was the mssql or oracle support. AFAIK `sequelize` is better.

PR #280, opened 2016-12-22
fix: call backend.submitPayment

sentientwaffle:
This will allow the backend to handle fees, making `transfer.depends_on` obsolete.

### DIFF
--- a/test/subscriptionSpec.js
+++ b/test/subscriptionSpec.js
@@ -82,13 +82,16 @@ describe('Subscriptions', function () {
     nock.cleanAll()
     this.clock.restore()
     process.env = _.cloneDeep(env)
+    this.wsCadLedger.close()
+    this.wsUsdLedger.close()
+    this.wsEurLedger.close()
+    this.wsCnyLedger.close()
   })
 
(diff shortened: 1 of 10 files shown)


test/subscriptionSpec.js
```diff
@@ -126,11 +126,19 @@ describe('Subscriptions', function () {
...
 
     sinon.assert.calledOnce(fulfillSpy)
     sinon.assert.calledWith(fulfillSpy, sourceId, 'cf:0:')
+    sinon.assert.calledOnce(backendSpy)
+    sinon.assert.calledWith(backendSpy, {
+      source_ledger: sourceTransfer.ledger,
+      source_amount: sourceTransfer.debits[0].amount,
+      destination_ledger: destinationTransfer.ledger,
+      destination_amount: destinationTransfer.credits[0].amount
+    })
```
emschwartz: I think it would be clearer to separate this into another unit test, rather than bundling all the things its supposed to do into one test

sentientwaffle: Ok; fixed

PR #288, opened 2017-01-05
Prepare removing ilp-core

### LINKED ISSUE #287: TradingPairs#addAll doesn't work if there is only one ledger
Adding a new plugin when there is only one current plugin will fail to populate the pairs array.

This is because of this code snippet:

https://github.com/interledgerjs/ilp-connector/blob/25449c109eb0b028357f18e7f9e72be28f1defa8/src/lib/trading-pairs.js#L20-L23

Since the pairs array will be empty if there is only one ledger, this code will fail to generate any new pairs.

justmoon:
Fixes #287.

Issue #287 was actually a somewhat tricky bug to fix, since the only thing right now that keeps track of the active ledgers is ilp-core. However, ilp-core doesn't know the currencies that the connector has assigned to each ledger (which can be different than the plugin's self-assigned currency), so even that doesn't have the information we need.

Since we are planning to remove ilp-core (interledgerjs/ilp-core#76), the connector will need to manage its ledger plugins itself. This refactor creates a class responsible for doing that. It currently still uses core internally, but creates a clear path for removing it once the quoting functionality is available as a separate module.


justmoon:
Not yet ready to merge - this PR makes a lot of changes and probably breaks a bunch of stuff. Planning to test it tomorrow and take it from there. #287 was preventing ILP kit peering from working so this work was necessary.

cc/ @vhpoet

justmoon:
Open issue: Currently we're passing the pairs to the rate backend - not a huge fan of that if the pairs are going to keep changing at runtime. To do this properly would require some kind of event system to notify the backend of the changed pairs, which seems complicated.

Is telling the backend about the pairs up-front actually an important use case? If not, we can just pass the currencies into `getQuote` which seems much simpler. (See fixerio backend in this PR.)

emschwartz:
@justmoon I think we wanted to replace `getQuote` with `getCurve`. Passing the currencies to either of those sounds fine to me though

justmoon:
Should be good to go now, please review.

src/lib/route-broadcaster.js
```diff
@@ -110,14 +108,18 @@ class RouteBroadcaster {
...
-    const connectors = yield client.getConnectors()
-    for (const connector of connectors) {
+    yield this._crawlLedgerPlugin(client.getPlugin())
+  }
+
+  * _crawlLedgerPlugin (plugin) {
+    const prefix = yield plugin.getPrefix()
+    const localAccount = yield plugin.getAccount()
+    const info = yield plugin.getInfo()
+    for (const connector of (info.connectors.map(c => c.name) || [])) {
```
sentientwaffle: I think the `|| []` is unnecessary, since `[].map(...)` is `[]`.

PR #289, opened 2017-01-05
feat: backends should have getCurve, not getQuote

sentientwaffle:


src/backends/fixerio/index.js
```diff
@@ -106,27 +104,10 @@ class FixerIoBackend {
...
-      destinationAmount = new BigNumber(params.source_amount).times(rate)
-    } else if (params.destination_amount) {
-      log.debug('creating quote with fixed destination amount')
-      sourceAmount = new BigNumber(params.destination_amount).div(rate)
-      destinationAmount = new BigNumber(params.destination_amount)
-    } else {
-      throw new NoAmountSpecifiedError('Must specify either source ' +
-        'or destination amount to get quote')
-    }
+    const sourceAmount = 100000000
```
justmoon: Would prefer if this was a named constant at the top of the file, e.g. `PROBE_SOURCE_AMOUNT` with a comment to explain what it's for and why it's a constant.

PR #291, opened 2017-01-09
fix: use setTimeout for route broadcasting

sentientwaffle:
Fixes https://github.com/interledgerjs/ilp-connector/issues/285

### DIFF
--- a/src/lib/route-broadcaster.js
+++ b/src/lib/route-broadcaster.js
@@ -53,30 +53,41 @@ class RouteBroadcaster {
         throw e
       }
     }
-    setInterval(() => this.routingTables.removeExpiredRoutes(), this.routeCleanupInterval)
-    defer.setInterval(function * () {
+    this.removeExpiredRoutesSoon()
+    this.broadcastSoon()
+  }
+
+  removeExpiredRoutesSoon () {
+    setTimeout(() => {
+      this.routingTables.removeExpiredRoutes()
+      this.removeExpiredRoutesSoon()
+    }, this.routeCleanupInterval)
+  }
+
+  broadcastSoon () {
+    defer.setTimeout(function * () {
       yield this.reloadLocalRoutes()
-      this.broadcast()
-    }.bind(this), this.routeBroadcastInterval)
+      yield this.broadcast()
+      this.broadcastSoon()
+    }, this.routeBroadcastInterval)
   }
 
   broadcast () {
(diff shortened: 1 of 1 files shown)


src/lib/route-broadcaster.js
```diff
@@ -55,30 +55,41 @@ class RouteBroadcaster {
         throw e
       }
     }
-    setInterval(() => this.routingTables.removeExpiredRoutes(), this.routeCleanupInterval)
-    defer.setInterval(function * () {
+    this.removeExpiredRoutesSoon()
+    this.broadcastSoon()
```
justmoon: Why are these two on separate timers?

src/lib/route-broadcaster.js
```diff
@@ -55,30 +55,41 @@ class RouteBroadcaster {
...
+      this.removeExpiredRoutesSoon()
+    }, this.routeCleanupInterval)
+  }
+
+  broadcastSoon () {
+    defer.setTimeout(function * () {
       yield this.reloadLocalRoutes()
-      this.broadcast()
-    }.bind(this), this.routeBroadcastInterval)
+      yield this.broadcast()
```
justmoon: You should probably use try/catch here and log any errors. (Same for the expired routes loop.)

For example, if `this.routingTables.toJSON` throws (line 67) even once, broadcasting would stop for good.

src/lib/route-broadcaster.js
sentientwaffle: They have different intervals (`routeCleanupInterval` vs `routeBroadcastInterval`)

sentientwaffle: Er, if I misinterpreted the question: I would imagine that the expiry interval would be several times smaller than the broadcast interval, because the former is much cheaper.

src/lib/route-broadcaster.js
```diff
@@ -55,30 +55,51 @@ class RouteBroadcaster {
...
       }
-    }
+      this.removeExpiredRoutesSoon()
+    }, this.routeCleanupInterval)
+  }
+
+  broadcastSoon () {
+    defer.setTimeout(function * () {
+      try {
+        yield this.reloadLocalRoutes()
```
justmoon: > I would imagine that the expiry interval would be several times smaller than the broadcast interval, because the former is much cheaper.

In that case, does it make sense to run `removeExpiredRoutes` in this loop as well? Otherwise, we might broadcast already expired routes, right?

PR #292, opened 2017-01-09
feat: add getPlugin

vhpoet:


sentientwaffle:
What is this going to be used for? I think `Client` is being phased out as part of https://github.com/interledgerjs/ilp-core/issues/76.

vhpoet:
@sentientwaffle I need a way to find out the trustline balances, with that I'll be able to do `getClient()->getBalance()`.

justmoon:
`getPlugin()` would also work for you, right? That's probably pretty future-proof wrt the ilp-core change.

PR #293, opened 2017-01-10
fix: fixerio: throw on getCurve if rates are undefined

vhpoet:
Just a suggestion, because the `BigNumber Error: new BigNumber() not a number: undefined` exception is super annoying.

### DIFF
--- a/src/backends/fixerio/index.js
+++ b/src/backends/fixerio/index.js
@@ -4,6 +4,7 @@ const _ = require('lodash')
 const request = require('co-request')
 const BigNumber = require('bignumber.js')
 const log = require('../../common').log.create('fixerio')
+const ServerError = require('five-bells-shared/errors/server-error')
 const healthStatus = require('../../common/health.js')
 
 const RATES_API = 'https://api.fixer.io/latest'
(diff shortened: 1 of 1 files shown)


sentientwaffle:
I don't understand how this would be triggered. `backend.connect()` is called before `getCurve` ever is (https://github.com/interledgerjs/ilp-connector/blob/cb839567a81c0ac47344eeaa160d9f592eccd562/src/app.js#L27-L38). Are you sure it's not trying to get the curve for currencies that fixerio doesn't know about?

vhpoet:
@sentientwaffle maybe not before it's connected but the `destinationRate` and `sourceRate` are undefined.

sentientwaffle:
Ok, in that case the error message `getCurve is called before the backend is connected` is inaccurate.

vhpoet:
@sentientwaffle how about now

sentientwaffle:
How about separate error messages for when `destinationRate` vs `sourceRate` are missing; something like `throw new ServerError('No rate available for currency ' + params.destination_currency)`.

vhpoet:
@sentientwaffle yep, much better. Updated.

PR #295, opened 2017-01-17
fix: use sync get{Prefix,Account,Info}

sentientwaffle:
Note: there is a minor change to the Backend API: the constructor now receives
a `getInfo` function instead of an `infoCache` object, since the latter no
longer exists.

Related to https://github.com/interledger/rfcs/pull/143

Depends on https://github.com/interledgerjs/ilp-core/pull/77 and https://github.com/interledgerjs/ilp-plugin-bells/pull/87


src/app.js
```diff
@@ -91,18 +90,13 @@ function createApp (config, ledgers, backend, routeBuilder, routeBroadcaster, ro
       currencyWithLedgerPairs: ledgers.getPairs(),
       backendUri: config.get('backendUri'),
       spread: config.get('fxSpread'),
-      infoCache: infoCache
+      getInfo: (ledger) => ledgers.getPlugin(ledger).getInfo()
```
emschwartz: Are you passing in the info or the function?

sentientwaffle: A function that retrieves the info.

gGololicic:
Read me should have the Backend description updated

PR #308, opened 2017-02-17
feat: BREAKING Use RejectionMessage for ILP Errors

sentientwaffle:
Depends on https://github.com/interledgerjs/ilp-plugin-bells/pull/98, https://github.com/interledgerjs/five-bells-shared/pull/179, https://github.com/interledgerjs/five-bells-integration-test/pull/63

### DIFF
--- a/src/errors/ilp-error.js
+++ b/src/errors/ilp-error.js
@@ -1,0 +1,12 @@
+'use strict'
+
+const BaseError = require('five-bells-shared').BaseError
+
+class IlpError extends BaseError {
+  constructor (rejectionMessage) {
+    super(rejectionMessage.message)
+    this.rejectionMessage = rejectionMessage
+  }
+}
+
+module.exports = IlpError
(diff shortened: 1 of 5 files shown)


src/lib/testPaymentExpiry.js
```diff
@@ -125,13 +145,21 @@ TransferTester.prototype.loadCaseExpiry = function * (caseID) {
     json: true
   })
   if (caseRes.statusCode !== 200) {
-    throw new UnacceptableExpiryError('Unexpected remote error: ' + caseRes.statusCode + ' ' + caseRes.body)
+    throw new IlpError({
+      code: 'T00',
```
emschwartz: Seems like a remote error should be treated differently. Shouldn't there be a way to say "I tried but the next guy didn't work"?

src/errors/ilp-error.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
+const BaseError = require('five-bells-shared').BaseError
```
emschwartz: Rather than extending all of them directly in the code, it would probably be better to have an ILP error module as suggested in https://github.com/interledgerjs/ilp-plugin-bells/issues/83

sentientwaffle:
Superseded by https://github.com/interledgerjs/ilp-connector/pull/322

PR #322, opened 2017-03-09
BREAKING RejectionMessage, new ilp-packet format

sentientwaffle:
PR includes https://github.com/interledgerjs/ilp-connector/pull/308

Depends on https://github.com/interledgerjs/five-bells-shared/pull/183, https://github.com/interledgerjs/ilp-core/pull/83, https://github.com/interledgerjs/ilp-plugin-bells/pull/109, https://github.com/interledgerjs/ilp-routing/pull/32, https://github.com/interledgerjs/five-bells-integration-test/pull/63


README.md
```diff
@@ -140,7 +140,7 @@ Breaking down that command:
 
 ## Payments
 
-The connector will facilitate an interledger payment upon receiving a notification for a transfer in which it is credited. That "source" transfer must have a `ilp_header` in its credit's `memo` that specifies the payment's destination and amount.
+The connector will facilitate an interledger payment upon receiving a notification for a transfer in which it is credited. That "source" transfer must have an `IlpHeader` in its credit's `memo.ilp` that specifies the payment's destination and amount.
```
emschwartz: Should be ILP Packet or ILP Payment Packet instead of ILP Header (we should make the naming consistent throughout, there are a couple more mentions of "header" in this PR too)

src/lib/route-builder.js
```diff
@@ -115,11 +109,23 @@ class RouteBuilder {
...
+        code: 'S01',
+        name: 'Invalid Packet',
+        message: 'source transfer is missing "ilp"'
+      })
+    }
+    let ilpHeader
+    try {
+      ilpHeader = packet.deserializeIlpPayment(Buffer.from(sourceTransfer.ilp, 'base64'))
+    } catch (err) {
+      throw new IlpError({
```
emschwartz: If it's not going to report the specific error that was caught it should at least log it so we can find the exact problem when debugging

README.md
sentientwaffle: I was referring to the `[IlpHeader.json](https://github.com/interledgerjs/five-bells-shared/blob/master/schemas/IlpHeader.json)` schema; is the naming not correct?

sentientwaffle:
@emschwartz updated

src/lib/route-builder.js
```diff
@@ -114,45 +108,64 @@ class RouteBuilder {
...
+    try {
+      ilpPacket = packet.deserializeIlpPayment(Buffer.from(sourceTransfer.ilp, 'base64'))
+    } catch (err) {
+      log.debug('error parsing ILP packet: ' + sourceTransfer.ilp)
+      throw new IlpError({
+        code: 'S01',
+        name: 'Invalid Packet',
+        message: 'source transfer has invalid ILP packet'
+      })
     }
```
emschwartz: Can you add a debug line to log the ilp packet's destination amount and account?

sentientwaffle:
@emschwartz updated

src/lib/route-builder.js
```diff
@@ -114,45 +108,67 @@ class RouteBuilder {
...
+      ilpPacket = packet.deserializeIlpPayment(Buffer.from(sourceTransfer.ilp, 'base64'))
+    } catch (err) {
+      log.debug('error parsing ILP packet: ' + sourceTransfer.ilp)
+      throw new IlpError({
+        code: 'S01',
+        name: 'Invalid Packet',
+        message: 'source transfer has invalid ILP packet'
+      })
+    }
+    log.debug('constructing destination transfer ' +
```
emschwartz: This is a little deceptive because that's the final destination, not the "destination transfer" as far as the connector is concerned

test/data/notificationSourceTransferPrepared.json
```diff
@@ -12,7 +12,7 @@
       "amount":"1.07",
       "account":"usd-ledger.mark",
       "memo":{
-        "ilp_header": {
+        "ilp": {
           "amount":"1.00",
```
emschwartz: Should this be an integer?

src/lib/route-builder.js
sentientwaffle: Hm, I could change `destinationAccount` to `finalAccount` (likewise for amount). Or `packetAccount`? What do you think?

emschwartz: You could just change it to "constructing transfer for ILP packet with account= and amount=" or something like that

sentientwaffle:
@emschwartz updated

PR #327, opened 2017-04-10
Update ilp-plugin-bells dependency

michielbdejong:
My theory is that the correctness of the ilp-connector -> ilp-plugin-bells dependency is only tested by the ilp-kit suite from the integration tests, and so tests are passing in this repo's master, but this is still incorrect, and is still making the mj-release-2.0 branch of ilp-kit fail (causing the `BigNumber Error: shift() argument not an integer: NaN` error, because `ilp-kit/node_modules/ilp-connector/node_modules/ilp-plugin-bells/src/lib/plugin.js` doesn't expose currencyScale.


michielbdejong:
This PR is coordinated with the following `mj-release-2.0` PRs:

* https://github.com/interledgerjs/ilp-connector/pull/327
* https://github.com/interledgerjs/five-bells-service-manager/pull/29
* https://github.com/interledgerjs/ilp-core/pull/84
* https://github.com/interledgerjs/five-bells-integration-test/pull/73

package.json
```diff
@@ -37,9 +37,9 @@
     "co-request": "^1.0.0",
     "eventemitter2": "^2.2.1",
     "five-bells-shared": "^23.0.0",
-    "ilp-core": "~13.0.0",
+    "ilp-core": "interledgerjs/ilp-core#mj-release-2.0",
```
dappelt: Should this be changed back to use the npm module?

michielbdejong: yeah, I think at merge time we should use the branch names in package.json, then:
* once things are merged, bump dependency versions on npm, e.g. ilp-core -> version 14
* then update the ilp-kit package.json
* then we can remove the branches from github

michielbdejong: bumped ilp-core's version to 14.0.0, and ilp-connector's version to 17.0.0. Will merge both so that npm modules get built.

PR #333, opened 2017-04-25
Don't broadcast peer routes

michielbdejong:
Implements https://github.com/interledgerjs/ilp-routing/issues/29

### DIFF
--- a/src/lib/message-router.js
+++ b/src/lib/message-router.js
@@ -8,6 +8,8 @@ const validate = require('./validate').validate
 const quoteModel = require('../models/quote')
 const log = require('../common/log').create('message-router')
 
+const PEER_LEDGER_PREFIX = 'peer.'
+
 /**
  * @param {Object} opts
  * @param {Config} opts.config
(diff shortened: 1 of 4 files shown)


src/lib/message-router.js
```diff
@@ -132,7 +134,8 @@ MessageRouter.prototype.receiveRoutes = function * (payload, sender) {
     // We received a route from another connector, but that connector
     // doesn't actually belong to the route, so ignore it.
     if (route.source_account !== sender) continue
-    if (this.routingTables.addRoute(route)) gotNewRoute = true
+    const isPeerRoute = route.destination_ledger.startsWith(PEER_LEDGER_PREFIX)
+    if (this.routingTables.addRoute(route) && !isPeerRoute) gotNewRoute = true
```
sentientwaffle: If RouteBroadcaster isn't going to broadcast the peer routes anyway, why is this needed?

michielbdejong: ah, good point. I'll switch them around so the lazy evaluation ensures that routes *to* `peer.` ledgers never enter routing tables.

PR #334, opened 2017-04-27
feat: include paths in route broadcasts

michielbdejong:
Companion to https://github.com/interledgerjs/ilp-routing/pull/40


src/lib/route-broadcaster.js
```diff
@@ -251,7 +251,8 @@ class RouteBroadcaster {
         [ connectorLedger, targetPrefix ],
         { minMessageWindow: this.minMessageWindow,
           sourceAccount: connector,
-          targetPrefix: targetPrefix }
+          targetPrefix: targetPrefix },
+        [ [] ]
```
sentientwaffle: Why specify this if it is already the default value?

src/lib/message-router.js
```diff
@@ -132,8 +132,10 @@ MessageRouter.prototype.receiveRoutes = function * (payload, sender) {
   let gotNewRoute = false
   for (const route of routes) {
     // We received a route from another connector, but that connector
-    // doesn't actually belong to the route, so ignore it.
+    // doesn't actually appear as the first hop of the route, so ignore it.
```
sentientwaffle: Why did the comment change? It is still checking whether or not the route belongs to the connector that sent it.

michielbdejong: What does 'connector belongs to a route' mean? Or is it 'route belongs to a connector'? I thought 'appear as the first hop of' seemed easier to understand, if I understood the original comment's meaning correctly.

sentientwaffle: ```if (route.source_account !== sender) continue```

`sender` is the address of the connector that forwarded us the route. If `route.source_account` doesn't match it, then we assume that the connector sent us a random route and we ignore it. 
In other words., we verify that the _route_ belongs to the _connector_ that broadcasted it (the code comment is backwards, though you could probably argue it either way).

michielbdejong: In particular, the restriction that it needs to be the *first* hop, right?

michielbdejong: OK, I'll change it to “We received a route from another connector, but that *route* doesn’t actually belong to the *connector*, so ignore it.”

PR #337, opened 2017-05-09
feat: liquidity routing

michielbdejong:
When prepending a trustline as a hop, clip the liquidity curve to the maximum inbound transfer, so `plugin.maxBalance()-plugin.getBalance()`. This should help avoid a lot of the failed payments we're currently seeing when trying to use ilp-kit to pay each other back for lunch.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -104,7 +104,8 @@ function createApp (config, ledgers, backend, routeBuilder, routeBroadcaster, ro
       currencyWithLedgerPairs: ledgers.getPairs(),
       backendUri: config.get('backendUri'),
       spread: config.get('fxSpread'),
-      getInfo: (ledger) => ledgers.getPlugin(ledger).getInfo()
+      getInfo: (ledger) => ledgers.getPlugin(ledger).getInfo(),
+      getBalance: (ledger) => ledgers.getPlugin(ledger).getBalance()
     })
   }
 
--- a/src/backends/fixerio/index.js
+++ b/src/backends/fixerio/index.js
@@ -8,7 +8,7 @@ const ServerError = require('five-bells-shared/errors/server-error')
 const healthStatus = require('../../common/health.js')
 // This simple backend uses a fixed (large) source amount and a rate to generate
 // the destination amount for the curve.
-const PROBE_SOURCE_AMOUNT = 100000000
+const PROBE_SOURCE_AMOUNT = 1000000000000
 
 const RATES_API = 'https://api.fixer.io/latest'
 
@@ -29,6 +29,7 @@ class FixerIoBackend {
 
     this.spread = opts.spread || 0
     this.getInfo = opts.getInfo
+    this.getBalance = opts.getBalance
     // this.ratesCacheTtl = opts.ratesCacheTtl || 24 * 3600000
 
     this.rates = {}
@@ -91,7 +92,7 @@ class FixerIoBackend {
    * @param {String} params.destination_ledger The URI of the destination ledger
    * @param {String} params.source_currency The source currency
    * @param {String} params.destination_currency The destination currency
-   * @returns {Object}
+   * @returns {Promise.<Object>}
    */
   * getCurve (params) {
     // Get ratio between currencies and apply spread
@@ -106,6 +107,9 @@ class FixerIoBackend {
       throw new ServerError('No rate available for currency ' + params.destination_currency)
     }
 
+    const sourceInfo = this.getInfo(params.source_ledger)
+    const destinationInfo = this.getInfo(params.destination_ledger)
+
     // The spread is subtracted from the rate when going in either direction,
     // so that the DestinationAmount always ends up being slightly less than
     // the (equivalent) SourceAmount -- regardless of which of the 2 is fixed:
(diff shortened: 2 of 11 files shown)


michielbdejong:
addition: not only check that the incoming balance would stay under maxBalance, but also that outgoing balance would stay above minBalance

src/app.js
```diff
@@ -104,7 +104,11 @@ function createApp (config, ledgers, backend, routeBuilder, routeBroadcaster, ro
       currencyWithLedgerPairs: ledgers.getPairs(),
       backendUri: config.get('backendUri'),
       spread: config.get('fxSpread'),
-      getInfo: (ledger) => ledgers.getPlugin(ledger).getInfo()
+      getInfo: (ledger) => {
```
sharafian: why not give the backends access to `ledgers`, if we're putting functions in there?

src/backends/fixerio/index.js
```diff
@@ -114,14 +117,21 @@ class FixerIoBackend {
...
-    const sourceScale = this.getInfo(params.source_ledger).currencyScale
-    const destinationScale = this.getInfo(params.destination_ledger).currencyScale
-    const sourceAmount = new BigNumber(PROBE_SOURCE_AMOUNT).shift(sourceScale)
-    const destinationAmount = new BigNumber(sourceAmount)
-      .times(rate)
-      .shift(destinationScale - sourceScale)
-    return { points: [[0, 0], [sourceAmount.toNumber(), destinationAmount.toNumber()]] }
+      .shift(destinationInfo.currencyScale - sourceInfo.currencyScale)
+
+    return (typeof sourceInfo.maxBalance === 'undefined'
```
sharafian: any reason to use typeof over just `=== undefined`?

src/backends/fixerio/index.js
```diff
@@ -114,14 +117,21 @@ class FixerIoBackend {
...
+      ? Promise.resolve()
+      : sourceInfo.getBalance()
+        .then(balance => [ sourceInfo.maxBalance - balance, Math.floor((sourceInfo.maxBalance - balance) * rate) ])
```
sharafian: Given that `getCurve` is a generator, I think it would be easier to understand if you `yield`ed the promise calls and used if statements instead of ternary expressions.

src/app.js
michielbdejong: I think giving access to the whole connected plugin is a bit ugly, the separation where only the info is accessible is a nice air gap.

I agree inserting the getBalance method into the info object is a bit ugly too. how about giving both `getInfo` and `getBalance`?

sharafian: that would make more sense

src/backends/ilp-quoter/index.js
```diff
@@ -114,7 +114,7 @@ class ILPQuoter {
       log.error('Error getting quote: ', JSON.stringify(result.body))
       throw new ServerError('Unable to get quote from backend.')
     }
-    return {
+    return Promise.resolve({
```
sharafian: getCurve is a generator function, so returning a promise is unnecessary.

src/backends/fixerio/index.js
```diff
@@ -114,14 +118,21 @@ class FixerIoBackend {
...
-    const sourceScale = this.getInfo(params.source_ledger).currencyScale
-    const destinationScale = this.getInfo(params.destination_ledger).currencyScale
-    const sourceAmount = new BigNumber(PROBE_SOURCE_AMOUNT).shift(sourceScale)
-    const destinationAmount = new BigNumber(sourceAmount)
-      .times(rate)
-      .shift(destinationScale - sourceScale)
-    return { points: [[0, 0], [sourceAmount.toNumber(), destinationAmount.toNumber()]] }
+      .shift(destinationInfo.currencyScale - sourceInfo.currencyScale)
+
+    return (sourceInfo.maxBalance === undefined
```
sharafian: getCurve is a generator, so can you restructure this to use yield, instead of returning a promise? As it's written here, yielding getCurve would return a promise, effectively making getCurve return a promise to a promise

michielbdejong: yes, sorry, just saw that comment, good idea.

michielbdejong:
I'll add a pessimistic margin, so that the liquidity curve stays valid if the balance changes by a tiny amount, and only changes if the balance changes by a big amount.

src/backends/fixerio/index.js
```diff
@@ -91,7 +92,7 @@ class FixerIoBackend {
    * @param {String} params.destination_ledger The URI of the destination ledger
    * @param {String} params.source_currency The source currency
    * @param {String} params.destination_currency The destination currency
-   * @returns {Object}
+   * @returns {Promise.<Object>}
```
michielbdejong: ah wait, it doesn't return a promise to an object, it yields an object. will change this back.

src/backends/one-to-one.js
```diff
@@ -46,23 +47,38 @@ class OneToOneBackend {
    * @param {String} params.destination_ledger The URI of the destination ledger
    * @param {String} params.source_currency The source currency
    * @param {String} params.destination_currency The destination currency
-   * @returns {Object}
+   * @returns {Promise.<Object>}
    */
   * getCurve (params) {
```
michielbdejong: basically the same changes as in the fixerio backend

test/mocks/mockPlugin.js
```diff
@@ -49,8 +49,8 @@ class MockPlugin extends EventEmitter {
 
   * _handleNotification () { }
 
-  * getBalance () {
-    return '123.456'
+  getBalance () {
+    return Promise.resolve('123.456')
```
michielbdejong: comply with the rfc :)

test/quoteSpec.js
```diff
@@ -266,7 +266,7 @@ describe('Quotes', function () {
         destination_ledger: 'usd-ledger.',
         destination_amount: '1056024', // EUR/USD Rate of 1.0592 - .2% spread - slippage
         destination_expiry_duration: '5',
-        liquidity_curve: [ [1000, 0], [1000000000000, 1057081598942.9185] ]
+        liquidity_curve: [ [999.9999999999999, 0], [1000000000000, 1057081598942.9186] ]
```
michielbdejong: these values changed slightly because I used numbers, not BigNumbers, to apply currency exchange rates

PR #345, opened 2017-05-10
feat: BREAKING request/response; ILQP packets

sentientwaffle:
* Depends on: {five-bells-shared,ilp{,-plugin-bells,-routing}}#dj-ilqp
* Add `connector.registerRequestHandler()` (used by ilp-kit to add some request handling).
* Use plugin request/response semantics.
* Use ILQP packets (supports quotes by source/destination amount, but not liquidity curve).


michielbdejong:
also related to https://github.com/interledger/rfcs/pull/198 and https://github.com/interledgerjs/ilp-plugin-virtual/pull/62 for request/response, and https://github.com/interledger/rfcs/blob/master/0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md#example-request should be updated as well as part of this refactor.

Also, depending on what we decide to do with https://github.com/interledgerjs/ilp-connector/issues/344#issuecomment-300619614, this refactor could require a "version 3" of ilp-kit, and then all nodes in the open Interledger network would need to update at the same time, coordinated with switching over connector.land to the new format. However, if we decide that remote quoting is only currently used for default routes, then technically the change is still breaking for ilp-kit, but the impact would be restricted to compatibility between two ilp-kits where one acts as the default route of the other, so that would not require a coordinated network upgrade.

src/lib/route-broadcaster.js
```diff
@@ -243,9 +242,10 @@ class RouteBroadcaster {
       const connector = configRoute.connectorAccount
       const targetPrefix = configRoute.targetPrefix
 
+      const maxUint = new BigNumber(2).pow(64).minus(1).toString()
```
emschwartz: This should just be a constant in here

src/lib/route-builder.js
```diff
@@ -144,32 +147,25 @@ class RouteBuilder {
       ilpPacket.account, ilpPacket.amount)
 
     const sourceLedger = sourceTransfer.ledger
-    // Use `findBestHopForSourceAmount` since the source amount includes the slippage.
     const nextHop = this.routingTables.findBestHopForSourceAmount(
       sourceLedger, ilpPacket.account, sourceTransfer.amount)
     if (!nextHop) {
       log.info('could not find route for source transfer: ' + JSON.stringify(sourceTransfer))
-      log.debug('current routing tables (simplified to 10 points): ' + JSON.stringify(this.routingTables.toJSON(10)))
```
emschwartz: Why are you taking this out?

michielbdejong:
I created mj-test-dj-ilqp branches on the following repos
* five-bells-shared
* five-bells-ledger <- no changes here, but it should use the changes in five-bells-shared
* five-bells-integration-test
* ilp-connector
* ilp-routing
* ilp-plugin-bells
* ilp-plugin-virtual
* ilp-kit <- no changes here, but it should use the changes in all components
* ilp

src/lib/route-builder.js
```diff
@@ -2,97 +2,100 @@
...
-const IlpError = require('../errors/ilp-error')
+const IncomingTransferError = require('../errors/incoming-transfer-error')
 const getDeterministicUuid = require('../lib/utils').getDeterministicUuid
 const log = require('../common/log').create('route-builder')
 const startsWith = require('lodash/startsWith')
 
 class RouteBuilder {
   /**
-   * @param {RoutingTables} routingTables
    * @param {Ledgers} ledgers
```
michielbdejong: add `@param quoter` here

src/lib/quoter.js
```diff
@@ -0,0 +1,116 @@
...
+   * @param {Integer} request.destinationHoldDuration
+   * @returns {Object}
+   */
+  * quoteByDestinationAmount (request) {
+    if (request.destinationAmount === '0') {
+      throw new InvalidAmountSpecifiedError('destinationAmount must be positive')
+    }
+    return yield this._quoteByAmount(request,
+      IlpPacket.serializeIlqpByDestinationRequest,
+      IlpPacket.deserializeIlqpByDestinationResponse)
```
michielbdejong: `* _quoteByAmount (request)` only takes one argument, not three.

src/lib/quoter.js
```diff
@@ -0,0 +1,116 @@
...
+   * @param {Integer} request.destinationHoldDuration
+   * @returns {Object}
+   */
+  * quoteBySourceAmount (request) {
+    if (request.sourceAmount === '0') {
+      throw new InvalidAmountSpecifiedError('sourceAmount must be positive')
+    }
+    return yield this._quoteByAmount(request,
+      IlpPacket.serializeIlqpBySourceRequest,
+      IlpPacket.deserializeIlqpBySourceResponse)
```
michielbdejong: `* _quoteByAmount (request)` only takes one argument, not three.

michielbdejong:
Do the unit tests test remote quoting? I'm running into ['MockPlugin.sendRequest is not implemented'](https://github.com/interledgerjs/ilp-connector/blob/1eab9ec5cc8e554eb167f2931080985f99df7237/test/mocks/mockPlugin.js#L32)

sentientwaffle:
@michielbdejong the unit tests pass for me. Can you send me the stack, and/or which test in particular is failing?

emschwartz:
@sentientwaffle @michielbdejong Should this be merged now? I think integration tests are failing because some of these changes have been merged but not all

michielbdejong:
Yes, I'll have a look.

PR #350, opened 2017-05-24
feat: configurable broadcastCurves and storeCurves

michielbdejong:


src/lib/route-builder.js
```diff
@@ -14,7 +14,6 @@ const startsWith = require('lodash/startsWith')
 class RouteBuilder {
   /**
    * @param {Ledgers} ledgers
-   * @param {Quoter} quoter
```
sentientwaffle: Why did this get removed? (maybe a rebase error)

michielbdejong: ah, yes. good catch

michielbdejong:
> ilp-connector/src/lib/quoter.js needs the changes that you applied to ilp-core -- dj-ilqp makes ilp-connector handles quoting w/o ilp-core

ah ok, thanks! I'll update that then.

michielbdejong:
@sentientwaffle can you review 06ed2aa1c6? My goal with this is that it should accept both the current format (`data` and JSON curves) and the new format for broadcast_route messages (`custom` and binary curves).

src/lib/quoter.js
```diff
@@ -52,9 +52,14 @@ class Quoter {
     const hop = this._findBestHopForAmount(request)
     if (!hop) return null
 
-    // If we know a local route to the destinationAccount, use the local route.
+    // note the confusion between `next`, `destination`, and `final`;
+    // `finalAmount` will be renamed to `destinationAmount` later in
+    // the hopToQuote function.
+    if (hop.sourceAmount && hop.finalAmount) {
```
sentientwaffle: If `hop.sourceAmount`/`hop.destinationAmount` is `0`, should it still return `hopToQuote()`?

michielbdejong: This line only checks that the routing table lookup was successful. If you were quoting by sourceAmount, and you found the finalAmount, you're done.  If you were quoting by finalAmount, and you found the sourceAmount, you're also done. 

Btw, we should really stop using destinationAmount in this way, can we rename it to sourceAmount ; next Amount ; finalAmount?

michielbdejong: To answer your question, if nextAmount is zero, then sourceAmount is lower than the transaction fee for the first hop. When this is the case, we should still just return that quoteResponse, to tell the user about it. It's no use to do a live (remote) quote.
If finalAmount is undefined, then it means we have a (default) route, but don't know what the curve is for it, so this if statement should not be entered, and it should fall through to the remote quoting code path below.

sentientwaffle: Given the curve `[ [0,0], [10, 5] ]`, if you quote by `sourceAmount=1`, the `finalAmount` would be `0` (it's actually `0.5`, but the destination amount gets rounded down).

michielbdejong: ah, right. so yes, then it should just return whatever the quote is, right?

sentientwaffle: If the `hop.sourceAmount`/`hop.finalAmount` is `0`, then yes, it should still return `hopToQuote(hop, destinationHoldDuration)`.

src/lib/config.js
```diff
@@ -181,6 +181,17 @@ function getLocalConfig () {
...
   const routeBroadcastEnabled =
     routeBroadcastEnabledString ? Config.castBool(routeBroadcastEnabledString) : true
+
+  // For a 'core' node in an open network, set both to true.
+  // For a 'periphery' node in an open network, set only the first one to true.
+  // For a node in a network where price competition between routes is not needed, set both to false.
+  const broadcastCurvesString = Config.getEnv(envPrefix, 'BROADCAST_CURVES')
+  const broadcastCurves =
+    broadcastCurvesString ? Config.castBool(broadcastCurvesString) : true
+  const storeCurvesString = Config.getEnv(envPrefix, 'STORE_CURVES')
```
sentientwaffle: The new environment variables should be documented in `ilp-connector`'s README.

test/quoteSpec.js
```diff
@@ -219,7 +219,7 @@ describe('Quotes', function () {
     }, 'eur-ledger.mary')
     this.config.routeBroadcastEnabled = true
 
-    this.ledgers.getPlugin('eur-ledger.').sendRequest = (request) => {
+    this.ledgers.getPlugin('cad-ledger.').sendRequest = (request) => {
```
sentientwaffle: Why did this change?

test/quoteSpec.js
```diff
@@ -282,17 +282,15 @@ describe('Quotes', function () {
...
     beforeEach(function * () {
-      const curve = new LiquidityCurve([ [0, 0], [10000, 20000] ]).toBuffer()
       this.config.routeBroadcastEnabled = false
       yield this.messageRouter.receiveRoutes({
         new_routes: [{
           source_ledger: 'eur-ledger.',
           destination_ledger: 'random-ledger.',
           min_message_window: 1,
-          source_account: 'eur-ledger.mary',
-          points: curve.toString('base64')
```
sentientwaffle: Removing points causes the test to fail with `InvalidBodyError: RoutingUpdate schema validation error: should have required property 'points'`. I think the `Routes` schema needs to make the field optional, for when `broadcastCurves = false`.

michielbdejong: yes, I created https://github.com/interledgerjs/five-bells-shared/commit/ab1ecb6b7688 for that, will continue working on it tomorrow.

michielbdejong:
The main question now is why this behavior changed:
https://github.com/interledgerjs/five-bells-integration-test/commit/9183e5701bd5cbff9c3b828fd01b024cd9682bd2

I guess the route is different based on curves than based on gratuity-routing, but would be good to understand the change and document it.

test/quoteSpec.js
michielbdejong: This was merged, `points` is now optional. Will tag versions of 5b-shared and ilp-routing so this PR can use them.

src/lib/config.js
michielbdejong: done

src/lib/quoter.js
michielbdejong: ok, afaict that's what I implemented now.

PR #352, opened 2017-05-30
fix: set transfer.to on outgoing transfers

emschwartz:
transfer.account has been deprecated

resolves https://github.com/interledgerjs/ilp-connector/issues/351


michielbdejong:
I don't think there are any plugins left that still rely on the `.account` field being present, so we might as well remove it altogether, and leave just the `.to` field?

PR #353, opened 2017-06-06
feat: use secret to generate transfer ids

emschwartz:
also allow the connector to be configured such that it will use the same transfer id for the destination as the source (even though this is not recommended)

### DIFF
--- a/src/lib/config.js
+++ b/src/lib/config.js
@@ -6,6 +6,7 @@ const Utils = require('../lib/utils')
 const _ = require('lodash')
 const logger = require('../common/log')
 const log = logger.create('config')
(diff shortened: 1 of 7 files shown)


src/lib/config.js
```diff
@@ -202,6 +203,15 @@ function getLocalConfig () {
   // }
   const ledgerCredentials = parseCredentials()
 
+  const secretString = Config.getEnv(envPrefix, 'SECRET')
```
sharafian: you can do all this as `const secret = Buffer.from(Config.getEnv(envPrefix, 'SECRET'), 'base64') || crypto.randomBytes(32)`

PR #354, opened 2017-06-14
[WiP] feat: stand-alone server

michielbdejong:


michielbdejong:
There are a number of things to consider here:
* for now, the stand-alone connector will be single-threaded
* plugin-virtual will use an in-memory cache of in-flight transactions, plus one key-value postgres table per trustline which holds the balance and the log of successful transactions
* plugin-bells relies on the ledger for all state
* there is no mechanism for rotating the logs, the admin should delete old transactions manually when the database gets too big
* peers can be configured by editing the "Peers" table in postgres; I can create something like:

```sh
./scripts/addPeer.sh hostname limit currency
```

scripts/server.js
```diff
@@ -0,0 +1,42 @@
...
+
+  if (!prefix) {
+    this.status = 422
+    this.body = JSON.stringify({ message: 'missing prefix' })
+    return
+  }
+
+  let plugin
+  try {
+    plugin = connector.getPlugin(prefix)
```
sharafian: should check for a valid auth token once it gets the plugin

michielbdejong:
We discussed this at length today. We did not reach a consensus, but my conclusion is that ilp-kit's API process already implements all the functionality of a stand-alone connector (configuration scripts, database schema, the way peering works), except for multi-process support.

IMHO we should not build a competing implementation, but even if we do implement a competing implementation, we know for sure that we want multi-process support, so we probably first want to:
* implement the PluginBackend interface, and make it use atomic postgres transactions
* make the rpc process somehow become aware of updates to the "Peers" list
* store routing tables in the storage backend

Once we've done all those things, you can separate ilp-kit's rpc endpoint from the other api end-points, or still decide to reimplement the database schemas and config variable names below the PluginBackend, but let's postpone this until we have worked out multi-threading support for the connector

michielbdejong:
Closing this for now, will concentrate on my other open PRs and on the postgres-based PluginBackend first.

PR #355, opened 2017-06-14
feat: BREAKING handle liquidity quote requests

sentientwaffle:
* Quote-by-source/destination uses liquidity quoting internally; payment is handled by cached quotes if available.
* Depends on interledgerjs/ilp-routing#46 and https://github.com/interledgerjs/ilp/pull/107


src/app.js
```diff
@@ -10,12 +10,13 @@ const loadConfig = require('./lib/config')
 const RoutingTables = require('./lib/routing-tables')
 const RouteBuilder = require('./lib/route-builder')
 const RouteBroadcaster = require('./lib/route-broadcaster')
+const CurveCache = require('./lib/curve-cache')
```
michielbdejong: why not use Route.curve as the curve cache?

src/lib/curve-cache.js
```diff
@@ -0,0 +1,84 @@
...
+    this.sources = {} // { sourceLedger ⇒ { appliesToPrefix ⇒ [liquidityQuote] } }
+    this.quoteExpiryDuration = params.quoteExpiry
+    this.quoteCleanupInterval = params.quoteCleanupInterval
+  }
+
+  start () {
+    setInterval(this.removeExpiredQuotes.bind(this), this.quoteCleanupInterval)
+  }
+
+  removeExpiredQuotes () {
```
michielbdejong: Why not just leave expired curves in the cache, and check their expiry date when you retrieve it?

src/app.js
sentientwaffle: `Route.curve` is the current curve. The `CurveCache` records liquidity-quotes until they have expired, so that they can be honored during payment.

src/lib/curve-cache.js
sentientwaffle: It does check the expiry at that time, but we still need to clean up the old curves occasionally.

src/app.js
michielbdejong: @emschwartz I think maybe we should discuss this in the first meeting of the routing working group? The data structures ilp-kit uses are an implementation detail, but if curves from quotes are kept longer/shorter than curves from routes, that does change the functionality. It feels to me like we haven't clearly defined the functional requirements of the connector, with respect to e.g. reacting to price changes. For instance, is the `expiresAt` field of a quote response an indication only, or should it always be honored even if market prices changed?

momerath42: Whether `expiresAt` is honored is ultimately a per-peering-relationship legal question.  My focus on route-selection heuristics is partly based on the assumption that games will be played with this.

michielbdejong: Okay, so in that case sometimes a node may want to remember what it has promised, but I'm not sure whether we should base ilp-kit's architecture on the need to support that.
Maybe to simplify this question; should liquiditycurves be stored:
a) where they are now, in the routing table (which is PrefixMap)
b) in the newly proposed CurveCache (choice of nextHop would stay in the routing table)
c) spread across both these two places, depending on how the liquidity curve was obtained

michielbdejong: My gut feeling is that c) would lead to overly complex code.

michielbdejong:
#350 was merged, so you can rebase on master. I also merged  interledgerjs/five-bells-shared#191 since #350 causes the same functional change.

src/lib/quoter.js
```diff
@@ -1,17 +1,88 @@
...
     this.tables = ledgers.tables
+    this.quoteExpiryDuration = config.quoteExpiry // milliseconds
+  }
+
+  /**
+   * @param {Object} request
+   * @param {IlpAddress} request.sourceAccount
+   * @param {IlpAddress} request.destinationAccount
+   * @param {Integer} request.destinationHoldDuration
+   * @param {Boolean} [request._checkCache] default: false
```
michielbdejong: What would this do when you set it to true? If this is not implemented yet, maybe add a code comment?

src/lib/config.js
```diff
@@ -18,6 +18,7 @@ const DEFAULT_SLIPPAGE = 0.001 // 0.1%
 const DEFAULT_ROUTE_BROADCAST_INTERVAL = 30 * 1000 // milliseconds
 const DEFAULT_ROUTE_CLEANUP_INTERVAL = 1000 // milliseconds
 const DEFAULT_ROUTE_EXPIRY = 45 * 1000 // milliseconds
+const DEFAULT_QUOTE_EXPIRY = 20 * 1000 // milliseconds
```
michielbdejong: I think default should be 45 because routes are broadcast only ever 30 seconds. So setting the quote expiry to 20 seconds would lead to remote quoting even where that's not desired

src/lib/quoter.js
```diff
@@ -1,17 +1,88 @@
...
+
+  * _quoteLiquidity (request) {
+    const hop = this.tables.localTables.findBestHopForSourceAmount(request.sourceAccount, request.destinationAccount, '0')
+    if (!hop) return Promise.resolve(null)
+    const connector = hop.bestHop
+    const fullRoute = hop.bestRoute
+    const shiftBy = request._shiftCurve
+      ? this.tables.getScaleAdjustment(this.ledgers, fullRoute.sourceLedger, fullRoute.nextLedger)
+      : 0
+    if (isCurveExpired(fullRoute)) {
```
michielbdejong: we need this logic for all three quote types, right? would it be an idea to always do a liquidity quote to deal with expired or missing curves, even if the user is asking about a specific amount?

src/lib/quoter.js
```diff
@@ -1,17 +1,88 @@
...
+      const headRoute = this.tables.localTables.getLocalRoute(fullRoute.sourceLedger, fullRoute.nextLedger)
+      const tailQuote = yield ILQP.quoteByConnector({
+        plugin: this.ledgers.getPlugin(fullRoute.nextLedger),
+        connector,
+        quoteQuery: {
+          destinationAccount: request.destinationAccount,
+          destinationHoldDuration: request.destinationHoldDuration
+        }
+      })
+      if (tailQuote.code) throw new IlpError(tailQuote)
```
michielbdejong: isn't there a cleaner way to check if the response is an error? something like `(tailQuote.responseType === ERROR)` instead of `(tailQuote.code)`?

src/lib/quoter.js
```diff
@@ -1,17 +1,88 @@
...
+      const tailCurve = new LiquidityCurve(tailQuote.liquidityCurve)
+      // The quote is more specific than the route.
+      if (fullRoute.targetPrefix.length < tailQuote.appliesToPrefix.length) {
+        log.warn('quote.appliesToPrefix="%s" is more specific than route.targetPrefix="%s"',
+          tailQuote.appliesToPrefix.length, fullRoute.targetPrefix.length)
```
michielbdejong: don't we still want to cache the curve in this case, just create a new (more specific) entry in the routing table?

src/lib/quoter.js
```diff
@@ -1,17 +1,88 @@
...
+      }
+      // The quote is more general than the route, so update the route.
+      fullRoute.curve = headRoute.curve.join(tailCurve)
+      fullRoute.curveExpiresAt = tailQuote.expiresAt.getTime()
+      fullRoute.minMessageWindow = (tailQuote.sourceHoldDuration - request.destinationHoldDuration) / 1000 + headRoute.minMessageWindow
```
michielbdejong: This only updates the route details in the quote response, right?

Don't we still want to store the curve in the routing table?

src/lib/quoter.js
michielbdejong: ah right, I see it's being used that way in `quoteBySourceAmount` below, great!

test/paymentsSpec.js
```diff
@@ -79,6 +79,13 @@ describe('Payments', function () {
 
     this.mockPlugin1 = this.ledgers.getPlugin('mock.test1.')
     this.mockPlugin2 = this.ledgers.getPlugin('mock.test2.')
+
+    // Populate the curve cache.
```
michielbdejong: Isn't this a local route? So it should already have a curve when it's added to the routing table?

test/routeBuilderSpec.js
```diff
@@ -91,6 +91,13 @@ describe('RouteBuilder', function () {
       secret: Buffer.from('VafuntVJRw6YzDTs4IgIU1IPJACywtgUUQJHh1u018w=', 'base64')
     })
     yield this.ledgers.connect()
+
+    // Populate the curve cache.
```
michielbdejong: Isn't this a local route? So it should already have a curve when it's added to the routing table?

test/routeBuilderSpec.js
```diff
@@ -237,6 +244,13 @@ describe('RouteBuilder', function () {
             })
           })
         }
+
+        // Populate the curve cache.
```
michielbdejong: I think this is not needed here, nor on [line 229/236 above](https://github.com/interledgerjs/ilp-connector/pull/355/files#diff-9b51c24e991fec48cf7254e7394a6843R236) since the route [already has a curve](https://github.com/interledgerjs/ilp-connector/pull/355/files#diff-9b51c24e991fec48cf7254e7394a6843R233)?

test/subscriptionSpec.js
```diff
@@ -81,6 +81,13 @@ describe('Subscriptions', function () {
 
     this.transferUsdPrepared = _.cloneDeep(require('./data/transferUsdPrepared.json'))
     this.transferEurProposed = _.cloneDeep(require('./data/transferEurProposed.json'))
+
+    // Populate the curve cache.
```
michielbdejong: not needed, since these are all local routes, right? they should already have curves in the routing table when they are created?

src/lib/quoter.js
michielbdejong: Right, as discussed on slack, this works because fullRoute is hop.bestRoute passed by reference.
But what if the prefix from the quote result is shorter (more generic) or longer (more specific) than the destinationLedger of fullRoute? Then we want to add a new route into the table, right?

README.md
```diff
@@ -121,6 +121,7 @@ npm start
 * `CONNECTOR_ROUTE_BROADCAST_INTERVAL` (default: `30000`) the frequency at which the connector broadcasts its routes to adjacent connectors.
 * `CONNECTOR_ROUTE_CLEANUP_INTERVAL` (default: `1000`) the frequency at which the connector checks for expired routes.
 * `CONNECTOR_ROUTE_EXPIRY` (default: `45000`) the maximum age of a route.
+* `CONNECTOR_QUOTE_EXPIRY` (default: `20000`) the maximum age of a quote.
```
michielbdejong: 45000 now

src/lib/quoter.js
```diff
@@ -1,17 +1,100 @@
...
+        }
+      })
+      if (tailQuote.responseType === IlpPacket.Type.TYPE_ILP_ERROR) {
+        throw new IlpError(tailQuote)
+      }
+      const tailCurve = new LiquidityCurve(tailQuote.liquidityCurve)
+      // The quote is more specific than the route.
+      if (fullRoute.targetPrefix.length < tailQuote.appliesToPrefix.length) {
+        this.localTables.addRoute({
+          source_ledger: fullRoute.nextLedger,
```
michielbdejong: this needs a code comment to explain why one route's next is another route's source.

src/lib/quoter.js
```diff
@@ -1,17 +1,100 @@
...
+          destination_ledger: tailQuote.appliesToPrefix,
+          source_account: connector,
+          min_message_window: (tailQuote.sourceHoldDuration - request.destinationHoldDuration) / 1000,
+          points: tailQuote.liquidityCurve
+        })
+        fullRoute = this.localTables.findBestHopForSourceAmount(
+          request.sourceAccount, request.destinationAccount, '0').bestRoute
```
michielbdejong: Directly accessing `.bestRoute` will break in this case: https://github.com/interledgerjs/ilp-routing/blob/master/src/lib/routing-table.js#L43 - can you be sure that will not happen on this code path?

src/lib/quoter.js
```diff
@@ -39,77 +125,44 @@ class Quoter {
...
-        destinationAccount: request.destinationAccount,
-        destinationHoldDuration
-      }
+  /**
+   * @param {IlpAddress} sourceLedger
+   * @param {IlpAddress} destination
+   * @param {Amount} sourceAmount
+   * @returns {Object}
+   */
+  * findBestPathForSourceAmount (sourceLedger, destination, sourceAmount) {
```
michielbdejong: It's maybe a bit confusing that path finding ('findBestPath...') calls liquidity quoting, which in turn ends up calling the  path finding code in the ilp-routing module. But I guess we just have that level of complexity now; we can maybe try to remove layers later in a refactor.

src/lib/quoter.js
```diff
@@ -39,77 +125,44 @@ class Quoter {
...
+    const quote = yield this.quoteBySourceAmount({
```
michielbdejong: This looks like paths it is going to trigger a quote request, which is not necessarily the case.
Maybe add a code comment explaining that this function ends up calling https://github.com/interledgerjs/ilp-routing/blob/master/src/lib/routing-table.js#L38 so the routing table is checked first, and remote quoting is only used if the route's curve is expired or missing.

src/lib/quoter.js
sentientwaffle: Yes, I think so. It is accessing the route `request.sourceAccount -> request.destinationAccount`, immediately after creating a route that will match that specification.

PR #357, opened 2017-06-26
feat: retry crawl if peer is down

michielbdejong:


michielbdejong:
You want the retries to happen in the background (the connector should already start broadcasting to the peers that were crawled so far),  but you're right, I'll see if I can find a cleaner way to achieve that. Maybe we should rename the function to 'startCrawlClient' or something, to make it clearer

src/lib/route-broadcaster.js
```diff
@@ -181,12 +182,21 @@ class RouteBroadcaster {
...
-    yield this._crawlLedgerPlugin(client.getPlugin())
+  * _startCrawlClient (client) {
+    try {
+      // wait for first attempt to succeed or fail
+      yield this._crawlLedgerPlugin(client.getPlugin())
+    } catch(e) {
+      // follow-up attempts happen in the background until successful
+      // but this generator function already returns
+      setTimeout(() => {
+        this._crawlClient(client)
```
sentientwaffle: Doesn't this function no longer exist?

michielbdejong: ah :) sorry. fixed.

src/lib/route-broadcaster.js
```diff
@@ -181,12 +182,21 @@ class RouteBroadcaster {
...
-    yield this._crawlLedgerPlugin(client.getPlugin())
+  * _startCrawlClient (client) {
+    try {
+      // wait for first attempt to succeed or fail
+      yield this._crawlLedgerPlugin(client.getPlugin())
+    } catch(e) {
+      // follow-up attempts happen in the background until successful
+      // but this generator function already returns
+      setTimeout(() => {
+        this._startCrawlClient(client)
```
sentientwaffle: Doesn't this need to be in a `co()`, since `_startCrawlClient` is a generator?

michielbdejong: I guess I should add a unit test to this PR :)

michielbdejong:
@sentientwaffle I vaguely remember you solved this issue (an underlying https://github.com/interledgerjs/ilp-kit/issues/435) in a different PR? If not I'll take another look at this once I'm back from holidays.

sentientwaffle:
I think you're right, https://github.com/interledgerjs/ilp-connector/pull/386 should fix this issue. When the plugin connects [`_crawlLedgerPlugin` is triggered](https://github.com/interledgerjs/ilp-connector/blob/7e4189de27640e02de336331ff6c4f6a617a6fed/src/lib/route-broadcaster.js#L187-L191).

michielbdejong:
ok, will close this then. we should test this in the next version of ilp-kit to check if it's really fixed now.

Issue #358, opened 2017-06-27
Adding a peer fails when connector balance is zero


michielbdejong:
Because of the new liquidity routing system (where the curve announced on a route displays is capped at the maximum amount that can be sent over that route), when you add a peer, the curve for the newly created incoming (delivery) route, has a curve with max y zero. This should actually be represented as a curve with zero points, and adding the peer should not fail. The curve can be updated later when the connector has some balance again on the local ledger.

Currently, this situation leads to a 500 error.

PR #359, opened 2017-06-27
fix: curve should be empty if connector is broke

michielbdejong:
attempt at fixing #358


michielbdejong:
I'll add a minBalance of 0 to the mockPlugin, so I can create a unit test that fails in master

michielbdejong:
Update the unit tests so that we can set a minBalance of 0 on the mockPlugin. With this, it should be easier to define a unit test for this PR; will continue work on this after my holidays.

PR #365, opened 2017-07-06
fix: remove ilp-core

sentientwaffle:


src/lib/ledgers.js
```diff
@@ -138,6 +140,32 @@ class Ledgers {
...
+
+  /**
+   * @param {IlpAddress} prefix
+   * @param {LedgerPlugin} plugin
+   */
+  addPlugin (prefix, plugin) {
+    if (prefix.slice(-1) !== '.') {
+      throw new Error('prefix must end with "."')
+    }
+    plugin.onAny(this._relayEvent)
```
sharafian: `onAny` is in EventEmitter2 but not in the built-in EventEmitter. Can this be implemented another way?

src/lib/ledgers.js
```diff
@@ -2,20 +2,29 @@
...
-    this._core = new ilpCore.Core({ routingTables })
     this._pairs = new TradingPairs()
     this._ledgers = new Map()
     this.requestHandler = co.wrap(this._requestHandler.bind(this))
+
+    const ledgers = this
+    this._relayEvent = function () {
+      const event = arguments[0]
+      const args = Array.prototype.slice.call(arguments, 1)
+      return ledgers.emitAsync.apply(ledgers, [event, this].concat(args))
```
sharafian: `emitAsync` is supported in EventEmitter2 but not EventEmitter. Is there any other way we can do this, or should we require plugins to expose the EventEmitter2 interface rather than the core node one?

PR #369, opened 2017-07-11
fix: adjust probeSourceAmount according to scales

sentientwaffle:
Fixes https://github.com/interledgerjs/ilp-routing/issues/50

### DIFF
--- a/src/backends/fixerio/index.js
+++ b/src/backends/fixerio/index.js
@@ -110,6 +110,11 @@ class FixerIoBackend {
     const sourceInfo = this.getInfo(params.source_ledger)
     const destinationInfo = this.getInfo(params.destination_ledger)
 
+    // Make sure that probeSourceAmount * rate doesn't exceed 15 significant digits.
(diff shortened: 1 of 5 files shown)


src/backends/fixerio/index.js
```diff
@@ -110,15 +110,19 @@ class FixerIoBackend {
     const sourceInfo = this.getInfo(params.source_ledger)
     const destinationInfo = this.getInfo(params.destination_ledger)
 
+    // Make sure that probeSourceAmount * rate doesn't exceed 15 significant digits.
+    const scaleDiff = destinationInfo.currencyScale - sourceInfo.currencyScale
+    const probeSourceAmount = scaleDiff > 0
+      ? PROBE_SOURCE_AMOUNT.shift(-scaleDiff)
+      : PROBE_SOURCE_AMOUNT
```
michielbdejong: what if both the source and the destination ledger have a course scale?

PR #371, opened 2017-07-13
fix app helper

michielbdejong:


michielbdejong:
maybe we should rename these functions to `getInfoForLedger` and `getBalanceForLedger`, for easier debugging when you see e.g. `this.getBalance` not found.

emschwartz:
Included in https://github.com/interledgerjs/ilp-connector/pull/372

PR #372, opened 2017-07-13
feat: allow config to override plugin.getInfo values

emschwartz:


src/backends/fixerio/index.js
```diff
@@ -122,15 +122,15 @@ class FixerIoBackend {
 
     let limit
     if (sourceInfo.maxBalance !== undefined) {
-      let balanceIn = parseInt(yield this.getBalance(params.source_ledger))
-      let maxAmountIn = sourceInfo.maxBalance - balanceIn
-      limit = [ maxAmountIn, maxAmountIn * rate ]
+      let balanceIn = yield this.getBalance(params.source_ledger)
+      let maxAmountIn = new BigNumber(sourceInfo.maxBalance).minus(balanceIn)
+      limit = [ maxAmountIn.toNumber(), maxAmountIn.times(rate).toNumber() ]
```
sharafian: Are you certain it's safe to use JS sized numbers here? These values could be up to 64 bits.

emschwartz: Nope, not sure, but the other values returned from this function are JS numbers

test/modifyPluginSpec.js
```diff
@@ -85,6 +85,30 @@ describe('Modify Plugins', function () {
...
+          currencyScale: 0
+        }
+      })
+
+      const info = this.ledgers._core.clients['eur-ledger-2.'].getPlugin().getInfo()
+      assert.equal(info.minBalance, '-10')
+      assert.equal(info.maxBalance, '10000')
+      assert.equal(info.prefix, 'test.other.prefix.')
+      assert.equal(info.currencyCode, 'XYZ')
+      assert.equal(info.currencyScale, 0)
```
michielbdejong: maybe avoid the double mention of each value here, maybe using [`assert.include(info, overrideInfo)`](http://chaijs.com/api/assert/#method_include) or even just `assert.deepEqual(info, overrideInfo)` probably (if you don't want `info` to have any fields that `overrideInfo` doesn't have)

michielbdejong: That's just a nit though, it's fine the way it is.

PR #380, opened 2017-07-19
chore: bump plugin bells

sharafian:


sharafian:
@sentientwaffle should this be a major version or patch?

PR #388, opened 2017-08-09
feat: resend routes to peer that rebooted

michielbdejong:


momerath42:
You'll have to change the RouteUpdate schema, but yeah, this should work

PR #390, opened 2017-08-23
fix: return T04: Insufficient Liquidity

sentientwaffle:
* Translate a plugin's `NotAcceptedError` on `sendTransfer()` to `T04: Insufficient Liquidity`.
* Also, "Insufficient Source Amount" is `R01`, not `R02`.

Related to https://github.com/interledgerjs/ilp-kit/issues/400

Depends on https://github.com/interledgerjs/five-bells-integration-test/pull/86 and https://github.com/interledgerjs/ilp-plugin-bells/pull/154

### DIFF
--- a/src/lib/route-builder.js
+++ b/src/lib/route-builder.js
@@ -2,6 +2,7 @@
 const _ = require('lodash')
 const BigNumber = require('bignumber.js')
 const packet = require('ilp-packet')
+const ilpErrors = require('./ilp-errors')
 const NoRouteFoundError = require('../errors/no-route-found-error')
 const UnacceptableExpiryError = require('../errors/unacceptable-expiry-error')
 const UnacceptableAmountError = require('../errors/unacceptable-amount-error')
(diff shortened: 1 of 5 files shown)


src/lib/route-builder.js
```diff
@@ -198,7 +198,7 @@ class RouteBuilder {
       const expectedFinalAmount = new BigNumber(ilpPacket.amount).times(1 - this.slippage)
       if (expectedFinalAmount.greaterThan(nextHop.finalAmount)) {
         throw new IncomingTransferError({
-          code: 'R02',
+          code: 'R01',
```
dappelt: > Also, "Insufficient Source Amount" is R01, not R02.

In order to avoid such bugs, how about using more expressive constants instead of hard-coding the error code? For example, in `ilp-packet` we could define something along the lines `const ERROR_R01_INSUFFICIENT_SOURCE_AMOUNT = {code: ... , name: ... }`

src/models/payments.js
```diff
@@ -25,6 +25,10 @@ function * settle (sourceTransfer, destinationTransfer, config, ledgers) {
           code: 'S00',
           name: 'Bad Request',
           message: 'destination transfer failed: ' + err.message
+        } : (err.name === 'NotAcceptedError') ? {
```
dappelt: If the plugin returns a `NotAcceptedError`, is that always due to `Insufficient Liquidity`? This is what RFC-0004 says about `NotAcceptedError`:
> An operation has been rejected due to ledger-side logic

I guess a `NotAcceptedError` can not always be translated into `Insufficient Liquidity`.

sentientwaffle: That is true. Plugins don't have an explicit "Insufficient Source Amount" error right now (it falls under `NotAcceptedError`) -- maybe they should. See: https://github.com/interledger/rfcs/pull/281

PR #394, opened 2017-11-07
Forwarded by should be an array

justmoon:
Tests currently fail, because of interledger/rfcs#332, see [RejectionMessage.json](https://github.com/interledgerjs/five-bells-shared/blob/96c77d5ad34efffd2e6d7004eb57cafb214a079c/schemas/RejectionMessage.json)


justmoon:
Closed, see: https://github.com/interledgerjs/five-bells-shared/pull/195#issuecomment-344402148

PR #396, opened 2017-11-21
Pass fulfillment data back

justmoon:
Pass fulfillment data, see interledger/rfcs#338.

### DIFF
--- a/test/subscriptionSpec.js
+++ b/test/subscriptionSpec.js
@@ -98,7 +98,8 @@ describe('Subscriptions', function () {
     const destinationTransfer = this.transferEurProposed
     const sourceAmount = '10700'
     const destinationAmount = '10000'
-
+    const fulfillment = 'HS8e5Ew02XKAglyus2dh2Ohabuqmy3HDM8EXMLz22ok'
+    const fulfillmentData = 'ABAB'
     const sendSpy = sinon.spy(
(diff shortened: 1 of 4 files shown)


test/subscriptionSpec.js
```diff
@@ -98,7 +98,8 @@ describe('Subscriptions', function () {
     const destinationTransfer = this.transferEurProposed
     const sourceAmount = '10700'
     const destinationAmount = '10000'
-
+    const fulfillment = 'HS8e5Ew02XKAglyus2dh2Ohabuqmy3HDM8EXMLz22ok'
+    const fulfillmentData = 'ABAB'
```
sentientwaffle: Why is `fulfillmentData` a base64-encoded string instead of a Buffer?

michielbdejong: That's how it comes from the LPI from the `outgoing_fulfill` event and also how it needs to be passed back into the LPI in the `fulfillCondition` call. There is no need to base64-unencode it and then reencoded it later. Note that the same is true for the `fulfillment` variable, `'HS8e...` is also a base64url-encoded buffer there. The relationship with the actual fulfillment data is as follows:

```js
const actualRawFulfillmentData = Buffer.from([ 0, 1, 2, 3])
const fulfillmentData = base64url(IlpPacket.serializeFulfillmentData(actualRawFulfillmentData))
```

PR #398, opened 2017-11-29
feat: use forwarded payments

michielbdejong:
Follow-up to #397. Work in progress, tests are still failing.


michielbdejong:
> should we just say it doesn't matter because it hasn't been in there for long and there isn't anything that's using that feature yet?

Yes, that was my plan. Specifically, even though it's been in master, no npm version was tagged in the meantime.

PR #399, opened 2017-12-05
fix: append account to forwardedBy array

emschwartz:
Related to https://github.com/interledger/rfcs/issues/332


emschwartz:
☹️  didn't see https://github.com/interledgerjs/ilp-connector/pull/394 before

PR #401, opened 2017-12-18
Update to ILPv4 + LPI2

justmoon:
Depends on interledgerjs/ilp-compat-plugin#7


package.json
```diff
@@ -31,13 +31,15 @@
...
-    "bignumber.js": "^3.0.1",
+    "bignumber.js": "^3.1.2",
     "co-request": "^1.0.0",
     "eventemitter2": "^2.2.1",
+    "extensible-error": "^1.0.2",
     "five-bells-shared": "^25.1.0",
     "ilp": "~11.2.0",
-    "ilp-packet": "~1.3.0",
+    "ilp-packet": "^2.0.0",
     "ilp-plugin-bells": "^15.0.0",
```
emschwartz: Should we stop including plugin bells?

src/lib/executeSourceTransfer.js
```diff
@@ -1,88 +0,0 @@
...
-  const timeoutPromise = new Promise((resolve, reject) => {
-    setTimeout(function () {
-      reject(`Fulfillment of source transfer ${sourceTransferID} timed out.`)
-    }, retryTimeout)
-  })
-
-  const fulfillPromise = promiseRetry(retryOpts, function (retry, number) {
-    return Promise.resolve(plugin.fulfillCondition(sourceTransferID, fulfillment, fulfillmentData))
-      .catch(function (err) {
-        if (shouldRetry(err)) {
```
emschwartz: Should we add fulfillment retry logic to the `ilp-compat-plugin`?

src/lib/utils.js
```diff
@@ -22,40 +20,6 @@ function getPairs (arr) {
   }, [])
 }
 
-/**
- * Deterministically generate a UUID from a secret and a public input.
```
emschwartz: Right now Mojaloop depends on having the same transfer ID across all of the ledgers. Taking the transfer id out of the LPI will make it more difficult to upgrade the connector used in that project. Do we have any alternative way of attaching the transfer ID from the previous ledger and making the connector use the same one on the outgoing transfer, or do we need to just make Mojaloop not depend on having the same transfer IDs?

src/lib/config.js
```diff
@@ -149,6 +149,8 @@ function getLocalConfig () {
   features.debugReplyNotifications =
     Config.castBool(Config.getEnv(envPrefix, 'DEBUG_REPLY_NOTIFICATIONS'))
 
+  const account = Config.getEnv(envPrefix, 'ILP_ADDRESS') || 'unknown'
```
sentientwaffle: Add `ILP_ADDRESS` to the README.

src/lib/route-broadcaster.js
```diff
@@ -39,9 +39,10 @@ class RouteBroadcaster {
 
     this.autoloadPeers = config.autoloadPeers
     this.defaultPeers = config.peers
-    // peersByLedger is stored in the form { ledgerPrefix ⇒ { connectorAddress ⇒ true } }
+    // peersByLedger is stored in the form { ledgerPrefix ⇒ true }
```
sentientwaffle: `s/peersByLedger/peers`

src/lib/route-broadcaster.js
```diff
@@ -133,95 +134,66 @@ class RouteBroadcaster {
...
-            new_routes: newRoutes,
-            hold_down_time: this.holdDownTime,
-            unreachable_through_me: unreachableLedgers,
-            request_full_table: requestFullTable
-          }
-        },
-        // timeout the plugin.sendRequest Promise just so we don't have it hanging around forever
-        timeout: this.routeBroadcastInterval
-      })
+    const broadcastPromise = this.ledgers.getPlugin(adjacentLedger).sendRequest({
```
michielbdejong: Does `this.ledgers.getPlugin(adjacentLedger)` return an LPI1 or an LPI2 interface? Because the `sendRequest` method, needed for broadcasting routes, seems to be [absent in LPI2](https://github.com/interledger/rfcs/blob/8ca2801402fa84f52b4399047911055c6200a09f/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md#methods)?

emschwartz: It's not in the spec but it is in the [`ilp-compat-plugin`](https://github.com/interledgerjs/ilp-compat-plugin/blob/master/index.js#L125-L142) as a deprecated feature

sentientwaffle: `sendRequest` won't exist when `compat()` returns the original LPI2 plugin (https://github.com/interledgerjs/ilp-compat-plugin/blob/d8331126a92ad2b65a7e7c4513ce44c13d485340/index.js#L31-L33).

michielbdejong: So then this PR will break route broadcasts?

src/lib/executeSourceTransfer.js
justmoon: I do think that fulfillment retries should be in the plugin because it depends on the ledger protocol if and how you want to retry.

Should it be in ilp-plugin-compat? I don't think it's worth the effort. Retry logic is an optimization, it's only to minimize unnecessary connector losses. I'd say we migrate to LPI2 and then add retry logic per plugin. Different plugins may want to handle retries quite differently, depending on whether they're TCP or UDP based, session or request based or to deal with some other ledger/transport-specific quirks.

src/lib/route-broadcaster.js
sentientwaffle: LPI1 plugins will work thanks to `compat`; LPI2 plugins will be broken.

michielbdejong: OK, but is there a plan to fix them, then? Or are we deprecating route broadcasts?

emschwartz: They should be switched to use either interledger payments or the internet

sharafian: in a similar way to @justmoon 's DHCP on ILP proposal (for asking your peer to assign you an ILP address), it would be easy to send a payment to your peer for which the destination is something like `peer.broadcast`, and the condition is either trivially fulfillable (if you're paying to send broadcasts) or unfulfillable (if you're not paying to send broadcasts). I don't see why that's worse/harder than using messaging.

michielbdejong: Sure, that's what plugin-xrp-escrow [already does](https://github.com/ripple/ilp-plugin-xrp-escrow/blob/master/src/plugin.js#L385). And if a BTP connection is available between the connectors then they can just use a BTP Message packet. But that's what `plugin.sendRequest` already nicely abstracts, and we agree that ilp-connector needs it, so then instead of removing code from the stack, we have just split the ledger plugin into a separate transfer plugin and messaging plugin, which is just moving code around.

sharafian: We don't actually need a messaging plugin anymore though, because transfers are expected to be just as fast as messages. And setting the amount to 1 means that the impact on your liquidity is negligible. This just lets us keep plugins as simple as possible, which should make them less buggy and more secure

michielbdejong: Ben and I discussed this on Slack and reduced this question to https://github.com/interledger/rfcs/issues/357

src/controllers/ildcp-host.js
```diff
@@ -0,0 +1,25 @@
...
+  async handle (sourceAccount, data) {
+    log.debug('responding to ILDCP config request. clientName=' + sourceAccount)
+
+    const info = this.accounts.getInfo(sourceAccount)
+
+    const writer = new Writer()
+    writer.writeVarOctetString(Buffer.from(sourceAccount, 'ascii'))
+    writer.writeUInt8(info.currencyScale)
+    writer.writeVarOctetString(Buffer.from(info.currency, 'utf8'))
+    return writer.getBuffer()
```
emschwartz: This encoding should probably be added to the `ilp-packet` module

emschwartz: This encoding should probably be added to the `ilp-packet` module

src/controllers/ilp-prepare.js
```diff
@@ -0,0 +1,108 @@
...
+      const { amount, executionCondition, destination } = parsedPacket
+
+      if (destination.startsWith(PEER_PROTOCOL_PREFIX)) {
+        return await this.peerProtocolController.handle(sourceAccount, parsedPacket)
+      }
+
+      const { nextHop, nextHopPacket } = await this.routeBuilder.getNextHopPacket(sourceAccount, parsedPacket)
+
+      log.debug('sending outbound ilp prepare. destination=%s amount=%s', destination, nextHopPacket.amount)
+      const result = await this.sendDataToPeer(nextHop, IlpPacket.serializeIlpPrepare(nextHopPacket))
```
emschwartz: Shouldn't we modify the ILP packet in place rather than re-encoding? (That could also be a separate PR though)

src/errors/remote-quote-error.js
```diff
@@ -0,0 +1,17 @@
...
+
+const { codes } = require('../lib/ilp-errors')
+
+class RemoteQuoteError extends BaseError {
+  constructor (message) {
+    super(message)
+
+    // TODO: Is this the right error code? Maybe we should pass on the error we
+    //   we received where possible?
+    this.ilpErrorCode = codes.T00_INTERNAL_ERROR
```
emschwartz: I'm not sure these errors will be temporary if other connectors just stop supporting ILQP

src/lib/plugin-store.js
```diff
@@ -1,12 +1,30 @@
...
 
 class PluginStore {
   constructor (uri, name) {
     this.name = name
-    this.db = new Database(uri)
+    const options = {
+      logging: log.debug,
+      omitNull: true,
+      // All transactions should be done with isolation level SERIALIZABLE
+      // TOOD:
```
emschwartz: What's this TODO for?

src/services/balances.js
```diff
@@ -0,0 +1,11 @@
+class Balances {
+  constructor (deps) {
+    this.balances = new Map()
+  }
+
+  handleMoney (amount) {
+    // TODO: Implement balance logic
```
emschwartz: This should be implemented ASAP after this PR is merged

src/services/route-builder.js
```diff
@@ -0,0 +1,359 @@
...
+const { getShortestUnambiguousPrefix } = require('../lib/utils')
+
+const PROBE_AMOUNT = new BigNumber(10).pow(14) // stays within 15 max digits for BigNumber from Number
+
+function rateToCurve (rate) {
+  rate = new BigNumber(rate)
+
+  // Make sure that neither amount exceeds 15 significant digits.
+  if (rate.gt(1)) {
+    return new LiquidityCurve([ [0, 0], [ PROBE_AMOUNT / rate, PROBE_AMOUNT ] ])
```
emschwartz: Shouldn't this be using bignumber.js' math functions?

src/controllers/ildcp-host.js
justmoon: Yeah, I wasn't sure where to put it. It's not ilp-packet layer, so I don't think ilp-packet is actually the right place unless we turn that into ilp-formats. Plus - is it really worth modularizing four lines of code?

src/controllers/ilp-prepare.js
justmoon: Imho, that's premature optimization. Not long ago, we were ok going with a text-based format that would have been an order of magnitude slower. It's just not worth worrying about trying to get zero-copy performance at this stage. There are much lower hanging fruit.

src/errors/remote-quote-error.js
justmoon: Good point - what's the right error code here? Or should we add one?

(I ran into a few cases where I was like: might be good to add a code for this - or make an existing code less specific.)

src/lib/plugin-store.js
justmoon: This "TOOD" was going to say: So we still need SERIALIZABLE isolation level. (My guess is no.)

src/services/balances.js
justmoon: Yeah. I think that's a job for @sharafian - he's written that code about 15 times already since it used to be in each plugin.

src/services/route-builder.js
justmoon: Copied the existing code to make sure the results are exactly the same - I still like that idea so we don't have to update the tests as much. But we'll see once we get to updating the integration tests how it'll play out.

src/controllers/ildcp-host.js
emschwartz: Fair enough

src/services/accounts.js
```diff
@@ -0,0 +1,201 @@
...
+
+    creds = _.cloneDeep(creds)
+
+    if (!this._config.validateAccount(accountAddress, creds)) {
+      throw new Error('error while adding account, see error log for details.')
+    }
+
+    const store = this.store.getPluginStore(accountAddress)
+
+    creds.options.prefix = accountAddress
```
sentientwaffle: Is this line necessary (see line 93 `prefix: accountAddress + '.',`)? If it is, shouldn't the prefix be set with a ` + '.'` suffix?

src/services/config.js
```diff
@@ -0,0 +1,285 @@
...
+    //  "peerAddress": "example.other."
+    // }]
+    this.routes = this.parseRoutes()
+
+    const features = this.features = {}
+    // Debug feature: Reply to websocket notifications
+    features.debugReplyNotifications =
+    Config.castBool(Config.getEnv(envPrefix, 'DEBUG_REPLY_NOTIFICATIONS'))
+
+    this.address = Config.getEnv(envPrefix, 'ILP_ADDRESS') || ''
```
sentientwaffle: This parameter should be documented in the README.

justmoon:
README needs a near total rewrite - not sure when I'll get around to tackling that. For now, if you want to know the new config format, consult `src/schemas/Config.json` please. It has types, descriptions and defaults for all the config variables.

src/services/accounts.js
justmoon: Removed

src/services/config.js
justmoon: Yeah, README needs a major rewrite...

src/routing/peer.js
```diff
@@ -0,0 +1,128 @@
...
+
+  getPrefix (prefix) {
+    return this.routes.get(prefix)
+  }
+
+  async broadcastRoutes ({ accounts, routes, broadcastCurves, holdDownTime, unreachableAccounts, requestFullTable = false, currentEpoch, timeout }) {
+    const newRoutes = routes.filter(route => route.epoch > this.epoch && route.nextHop !== this.address).map(route => ({
+      source_ledger: this.address,
+      destination_ledger: route.prefix,
+      min_message_window: 1,
```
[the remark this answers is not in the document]
justmoon: Removed for now.

src/controllers/ccp.js
```diff
@@ -0,0 +1,39 @@
...
+    log.debug('received routes. sender=%s', sourceAccount)
+
+    const routeUpdate = {
+      newRoutes: payload.new_routes.map(route => (
+        route.source_ledger !== route.source_account
+          ? null
+          : {
+            peer: sourceAccount,
+            prefix: route.target_prefix || route.destination_ledger,
+            distance: Math.max((route.paths && route.paths[0] && route.paths[0].length) || 1, 1),
```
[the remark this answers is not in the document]
justmoon: Removed due to other refactors.

src/backends/fixerio/index.js
```diff
@@ -111,46 +110,22 @@ class FixerIoBackend {
...
    * Dummy function because we're not actually going
    * to submit the payment to any real backend, we're
-   * just going to execute it on the ledgers we're connected to
+   * just going to execute it on the accounts we're connected to
    *
-   * @param {String} params.source_ledger The URI of the source ledger
-   * @param {String} params.destination_ledger The URI of the destination ledger
-   * @param {String} params.source_amount The amount of the source asset we want to send
-   * @param {String} params.destination_amount The amount of the destination asset we want to send
+   * @param {String} params.sourceAccount The URI of the source ledger
```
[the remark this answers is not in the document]
justmoon: Not anymore after config refactor.

src/backends/fixerio.ts
```diff
@@ -0,0 +1,134 @@
...
+  protected rates: {
+    [key: string]: number
+  }
+  protected currencies: string[]
+
+  /**
+   * Constructor.
+   *
+   * @param {Integer} opts.spread The spread we will use to mark up the FX rates
+   * @param {String} opts.ratesApiUrl The API endpoint we will request rates from
```
emschwartz: The jsdoc is out of date. We should either update it or remove it if it's no longer necessary (since we're switching to typescript)

src/backends/one-to-one.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Nothing to do since this backend is totally static.
+   */
+  async connect () {
+    // Nothing to do
+  }
+
+  /**
+   * Get a liquidity curve for the given parameters.
```
emschwartz: This jsdoc is also out of date, should probably be removed

src/controllers/ildcp-host.ts
```diff
@@ -0,0 +1,28 @@
...
+    const peerAddress = this.accounts.getChildAddress(sourceAccount)
+    log.debug('responding to ILDCP config request. clientAddress=' + peerAddress)
+
+    const info = this.accounts.getInfo(sourceAccount)
+
+    const writer = new Writer()
+    writer.writeVarOctetString(Buffer.from(peerAddress, 'ascii'))
+    writer.writeUInt8(info.assetScale)
+    writer.writeVarOctetString(Buffer.from(info.assetCode, 'utf8'))
+    return writer.getBuffer()
```
emschwartz: Doesn't make a huge difference but this can now use https://github.com/interledgerjs/ilp-protocol-ildcp

src/controllers/ilp-prepare.ts
```diff
@@ -0,0 +1,100 @@
...
+          const errInfo = (err && typeof err === 'object' && err.stack) ? err.stack : err
+          log.warn('sending money to peer failed. peerId=%s amount=%s error=%s', nextHop, nextHopPacket.amount, errInfo)
+        })
+
+      this.backend.submitPayment({
+        sourceAccount: sourceAccount,
+        sourceAmount: amount,
+        destinationAccount: nextHop,
+        destinationAmount: nextHopPacket.amount
+      })
```
emschwartz: Maybe log if this has an error too?

src/controllers/ilp-prepare.ts
```diff
@@ -0,0 +1,100 @@
...
+  constructor (deps: reduct.Injector) {
+    this.config = deps(Config)
+    this.accounts = deps(Accounts)
+    this.routeBuilder = deps(RouteBuilder)
+    this.backend = deps(RateBackend)
+    this.peerProtocolController = deps(PeerProtocolController)
+  }
+
+  async handle (sourceAccount: string, packet: Buffer) {
+    log.debug('handling ilp prepare. sourceAccount=%s', sourceAccount)
```
emschwartz: Might be worth logging the whole packet?

src/controllers/peer-protocol.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.ildcpHostController = deps(IldcpHostController)
+  }
+
+  async handle (sourceAccount: string, parsedPacket: IlpPacket.IlpPrepare) {
+    if (!PEER_PROTOCOL_CONDITION.equals(parsedPacket.executionCondition)) {
+      throw new InvalidPacketError('condition must be null.')
+    }
+
+    switch (parsedPacket.destination) {
+      case 'peer.config':
```
emschwartz: Nitpick: put `'peer.config'` as a constant up at the top of the file

src/errors/no-related-source-credit-error.js
```diff
@@ -1,18 +0,0 @@
-'use strict'
-
-const UnprocessableEntityError =
-require('five-bells-shared/errors/unprocessable-entity-error')
-
-class NoRelatedSourceCreditError extends UnprocessableEntityError {
-
-  * handler (ctx, log) {
-    log.warn('No Related Source Credit: ' + this.message)
```
emschwartz: lol that this was still in there

src/errors/remote-quote-error.ts
```diff
@@ -0,0 +1,17 @@
...
+const { codes } = require('../lib/ilp-errors')
+
+export default class RemoteQuoteError extends BaseError {
+  public ilpErrorCode: string
+
+  constructor (message: string) {
+    super(message)
+
+    // TODO: Is this the right error code? Maybe we should pass on the error we
+    //   we received where possible?
```
emschwartz: This sounds like a good idea

src/lib/ilp-errors.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import * as IlpPacket from 'ilp-packet'
+
+export interface RejectionInfo {
+  code: string,
+  message: string,
+  data: Buffer
+}
+
+export const createIlpRejection = (account: string, { code, message, data }: RejectionInfo) => {
```
emschwartz: Nit: technically this should be called `createIlpReject`

src/lib/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+ * "a" because there may be a more specific route like "a.c".
+ *
+ * So we would call this utility function to find out that the least specific
+ * prefix for which there are no other more specific routes is "a.b".
+ *
+ * In order to force a minimum prefix, it can be passed as the third parameter.
+ * This function may make it even more specific if necessary to make it
+ * unambiguous, but it will never return a less specific prefix.
+ */
+export const getShortestUnambiguousPrefix = <T> (prefixMap: PrefixMap<T>, address: string, prefix = '') => {
```
emschwartz: Shouldn't this method go in the `prefix-map` file, or at least in the routing directory?

src/schemas/Config.ts
```diff
@@ -0,0 +1,133 @@
...
+   */
+  accounts: {
+    /**
+     * Description of individual account.
+     */
+    [k: string]: {
+      /**
+       * Relationship between the connector and the counterparty that the account is with.
+       */
+      relation: "parent" | "peer" | "child";
```
emschwartz: Do you think it would be more or less clear to adopt IP's terminology of upstream/downstream?

src/schemas/Config.ts
```diff
@@ -0,0 +1,133 @@
...
+      };
+      /**
+       * What segment will be appended to the connector's ILP address to form this account's ILP address. Only applicable to accounts with `relation=child`. Defaults to the id of the account, i.e. the key used in the `accounts` config object.
+       */
+      ilpAddressSegment?: string;
+      [k: string]: any;
+    };
+  };
+  /**
+   * Which accounts should be use as peers for routing purposes. Defaults to the set of accounts set to `relation=peer`.
```
emschwartz: Why would this ever be different than the set of accounts listed as peers? Also, it might be useful to explain a bit more what "use[d] as peers for routing purposes" means (also note the typo)

src/schemas/Config.ts
```diff
@@ -0,0 +1,133 @@
...
+   */
+  peers?: string[];
+  /**
+   * Additional routes to add to the connector's routing table.
+   */
+  routes?: {
+    /**
+     * ILP address prefix that this route applies to. Configured routes take precedence over the same or shorter prefixes that are local or published by peers. More specific prefixes will still take precedence.
```
emschwartz: The fact that more specific prefixes always take precedence seems potentially problematic. If I want to configure my connector to always send `g.crypto.xrp.` to you, I don't want a route broadcast to be able to override that just by including a long list of specific XRP addresses.

src/types/backend.ts
```diff
@@ -0,0 +1,12 @@
+export interface SubmitPaymentParams {
+  sourceAccount: string
+  destinationAccount: string
+  sourceAmount: string
+  destinationAmount: string
+}
+
+export interface IBackend {
```
emschwartz: What does `IBackend` mean?

src/types/plugin.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface DataHandler {
+  (data: Buffer): Promise<Buffer>
+}
+
+export interface MoneyHandler {
+  (amount: string): Promise<void>
+}
+
+export interface IPlugin {
```
emschwartz: Note this is also exported from `ilp-compat-plugin` ([here](https://github.com/interledgerjs/ilp-compat-plugin/blob/master/src/index.ts#L49))

test/config.test.js
```diff
@@ -0,0 +1,208 @@
...
+const logHelper = require('./helpers/log')
+const env = _.cloneDeep(process.env)
+
+describe('Config', function () {
+  logHelper(logger)
+
+  describe('parseConnectorConfig', function () {
+    beforeEach(function () {
+      process.env.CONNECTOR_ACCOUNTS = JSON.stringify({
+        'usd-ledger': {
```
sentientwaffle: Are these ledger prefixes? Shouldn't it be `"usd-ledger."` (with a trailing `.`)?

justmoon: No, the keys on the accounts map are now just some arbitrary names ("account id") that you assign. Generally, the only thing they're used for now is for referring to the account. The reason is that ILP addresses are ephemeral, so not a good idea to hard-code them in the config.

The one exception is child ledgers. The connector will by default assign the address `[own address].[child's account id]` to the child. You can override that with the `ilpAddressSegment` option in the account options.

README.md
```diff
@@ -124,9 +104,9 @@ npm start
 * `CONNECTOR_QUOTE_EXPIRY` (default: `45000`) the maximum age of a quote.
 * `CONNECTOR_BACKEND` (default: `'fixerio'`) the backend used to determine rates. This can either be a module name from `src/backends/` or a different module that will be `require()`ed by the connector.
 * `CONNECTOR_SECRET` (default: 32 random bytes) base64url-encoded 32-byte secret used by the connector to deterministically randomize transfer IDs
-* `CONNECTOR_UNWISE_USE_SAME_TRANSFER_ID` (default: false, obviously) use the same transfer id for the destination transfer as the source transfer. If ledgers enforce transfer ID uniqueness (as they SHOULD), this means that someone can squat on an ID on the destination ledger and make this connector seem unreliable
 * `CONNECTOR_BROADCAST_CURVES` (default: true) include the liquidity curve when broadcasting a route
 * `CONNECTOR_STORE_CURVES` (default: true) store liquidity curves from route broadcasts (current implementation only stores them in-memory)
+* `CONNECTOR_REFLECT_PAYMENTS` (default: true) whether the connector is willing to send payments back to the sender if it thinks that is the best route
```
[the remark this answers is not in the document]
justmoon: Basically, if you receive a payment from peer `x`, this parameter defines where you are willing to send it back to `x` if that seems like the best route.

Useful if you have a multi-user plugin like mini-accounts and you want it to route all of its payments through the main connector, for instance, to let that connector apply your standard fee.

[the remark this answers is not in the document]
justmoon: I'll make sure to explain that better after the README rewrite.

src/backends/fixerio.ts
justmoon: Updated to tsdoc in next commit.

README.md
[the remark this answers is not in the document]
emschwartz: Oh, so the payment goes `Plugin P` -> `Connector` -> `Plugin P` just to apply the connector's spread? Is there a reason you wouldn't set that to true? If you set it to false would it reject the payment or send it to a different plugin?

src/controllers/peer-protocol.ts
justmoon: I think we should keep it as a literal. The reason we usually try to name constants is because this:

``` js
car.setType(3)
```

is a lot harder to understand than:

``` js
car.setType(TYPE_MODEL_S)
```

But when the constant is a meaningful string like `peer.config`, it's only used in one place in the file and it's clear from the context that it is the distinguisher for detecting ILDCP, then a constant just adds an unnecessary level of indirection imo.

src/lib/ilp-errors.ts
justmoon: Oops, that was actually dead code. Removed.

src/schemas/Config.ts
justmoon: Removed `CONNECTOR_PEERS` and added more easily understandable `sendRoutes` and `receiveRoutes` boolean options on each account.

src/schemas/Config.ts
justmoon: Yes, this is a problem with the current routing protocol and will be fixed when we introduce authenticated routes.

src/types/backend.ts
justmoon: Interfaces are commonly prefixed with `I` in many object-oriented languages, particularly .NET. I guess .NET developers brought the practice to TypeScript.

You can [go here](https://github.com/Microsoft/TypeScript-Handbook/issues/121) to read a bunch of people debating the pros and cons of prefixing interfaces with `I`.

I'll get rid of the `I`, there isn't any strong reason we need it and if it's confusing for you, it's probably going to be confusing for others.

src/types/plugin.ts
justmoon: Good point, but I think it's worth for the connector to have its own definition. ilp-compat-plugin shouldn't be used for anything other than converting old plugins I think.

README.md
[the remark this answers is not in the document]
justmoon: > If you set it to false would it reject the payment or send it to a different plugin?

If you set it to false, it would reject with `no route found`.

> Is there a reason you wouldn't set that to true?

Maybe to prevent some kind of infinite loop? Not sure. But `false` was the old behavior, so I wanted there to be a way to get the old behavior back if it's needed for any reason.

src/types/accounts.ts
```diff
@@ -1,4 +1,10 @@
 export interface AccountInfo {
+  relation: 'parent' | 'peer' | 'child',
```
emschwartz: Still think we should use `upstream`, `peer`, and `downstream` instead of `parent` / `child`. Parent makes me think of something that spawns other things. Big connectors don't spawn smaller ones, the smaller ones might be customers of the big one but they're still independent. What we're talking about is much more related to ISPs being "upstream" or "downstream" from one another than something like class inheritance.

sharafian: In ISP terms, you could have an upstream peer, a downstream peer, or a peer which is not strictly either. This could cause some confusion because those terms wouldn't be mutually exclusive.

Peer, Parent, and Child are mutually exclusive, and also describe relationships between people which are easy to conceptualize. In hierarchical descriptions, Parent and Child have easy to grasp meanings, so adding the definition of Peer as a side-to-side link isn't hard to conceptualize.

src/common/index.ts
```diff
@@ -0,0 +1,5 @@
+import * as log from './log'
```
sentientwaffle: Is `src/common/index.ts` file actually required anywhere? Maybe the logger could move to services.

michielbdejong:
Trying to run `npm install` on this branch, I get `npm ERR! 404  '@justmoon/json-schema-to-markdown' is not in the npm registry.`

sharafian:
Could that be an npm 5 issue? Are you on node 8?

justmoon:
My bad, when you upload namespaced packages on NPM you have to explicitly make them public. Should be fixed now.

src/lib/route-builder.js
```diff
@@ -1,272 +0,0 @@
...
-    if (!nextHop) {
-      log.info('could not find quote for source transfer: ' + JSON.stringify(sourceTransfer))
-      throw new IncomingTransferError(ilpErrors.F02_Unreachable({
-        message: 'No route found from: ' + sourceLedger + ' to: ' + ilpPacket.data.account
-      }))
-    }
-    this._verifyLedgerIsConnected(nextHop.destinationLedger)
-
-    // As long as the fxSpread > slippage, the connector won't lose money.
-    const expectedSourceAmount = new BigNumber(nextHop.sourceAmount).times(1 - this.slippage)
```
sentientwaffle: This slippage computation wasn't added back in the refactor (the config option still exists, though).

README.md
```diff
@@ -321,6 +341,29 @@ Shorthand for `config.storeConfig.path`.
...
+
+* Environment: `CONNECTOR_MIDDLEWARES`
+* Type: `array`
+* Default: `[]`
+
+| Name           | Type    | Description                                                              |
+| -------------- | ------- | ------------------------------------------------------------------------ |
+| `[]`           | object  | Object describing middleware instance.                                   |
+| `[].type`      | string  | NPM module that should be `require`d to load the middleware constructor. |
+| `[].priority`  | integer | Priority at which this middleware should be inserted. Must be unique.    |
```
sentientwaffle: Is this field implemented? I don't see it used in the code.

justmoon: Nope, it was from a previous version, I'll take it out!

src/lib/route-builder.js
justmoon: Removed.

README.md
justmoon: Removed.

emschwartz:
Woohoo! What a monster PR. Nicely work @justmoon

PR #402, opened 2018-01-10
bob and alice example (LPI v1 and v2)

michielbdejong:


michielbdejong:
The V8 version used is actually still v6 ;)

michielbdejong:
Adding a simple example that plugs two dummy V2 plugins into the connector (alice and bob) and completes a conditional transfer from alice to bob. Right now, what i'm running into is that it seems child plugins don't show up in the routing table, so i'll read up on IL-DCP and decide whether to make bob a parent so that he can choose his own ILP address, or if Bob can ask the connector what his ILP address is (not `test.quickstart.bob`, apparently)

sharafian:
If this is creating a connector from within node, it should probably use the connector's constructor instead of modifying `process.env`

michielbdejong:
One thing that's still ugly is that it tells the connector to create the plugins by [requiring `../../examples/helpers/peerPlugin` as a relative path](https://github.com/interledgerjs/ilp-connector/pull/402/files#diff-9e07fc613df8a4ec321f4bbd3835e6bbR14), the number of `..` would have to change if the code that instantiates plugins gets moved inside this repo.

michielbdejong:
thanks!

PR #419, opened 2018-02-28
fix: use correct error codes

sentientwaffle:
Also adds `F08: Maximum Payment Size Exceeded`.

Fixes https://github.com/interledgerjs/ilp-connector/issues/417

See https://github.com/interledger/rfcs/issues/397

Tests should pass once https://github.com/interledgerjs/ilp-connector/pull/416 is merged.


justmoon:
> The connector should use the errors added to the ilp-packet module in interledgerjs/ilp-packet#34

Agreed, but that can be a separate PR.

PR #426, opened 2018-04-14
chore: remove git dependencies for mirror plugin

justmoon:
Removed LPIv1 example. (LPIv1 is deprecated now.)


michielbdejong:
Good point about removing that dev dependency.

The only reason I still maintain ilp-plugin-mirror-v1 is that amundsen still [wraps all of ilp-connector into an LPIv1 plugin](https://github.com/interledgerjs/amundsen/blob/master/src/bmp.js#L126) so that it can integrate with the older parts of Amundsen, but that's specifically because Amundsen's goal is to support the last four or six quarterly versions of our stack.

PR #431, opened 2018-05-06
Performance: Cleanup redundant routes

justmoon:
Previously, connectors would broadcast a full history of their routing table changes:

```
1: g.u1 => g.u1
2: g.u2 => g.u1 g.u2
3: g.u2 withdrawn
```

However, if an update is overwritten by a later update, we can safely delete it. E.g. this log is equivalent:

```
1: g.u1 => g.u1
3: g.u2 withdrawn
```

Note that we can't delete update no.3 even though it is now a no-op for anyone downloading the table from scratch. This is because someone might have downloaded our routing table up to epoch 2 and they still need to know that we no longer serve this route.

Both route changes and route withdrawals can be removed from the log when a new entry for the same entry is added, regardless of whether the new entry is adding, changing, or removing the route. When adding a route update to the log, there will be at most one previous update for that prefix in the log. We may safely delete it.

### DIFF
--- a/src/routing/dragon.ts
+++ b/src/routing/dragon.ts
@@ -1,9 +1,8 @@
 import { Route } from '../types/routing'
-import PrefixMap from './prefix-map'
 import { create as createLogger } from '../common/log'
 const log = createLogger('dragon')
 import { Relation, getRelationPriority } from './relation'
-import ForwardingRoutingTable from '../services/forwarding-routing-table';
+import ForwardingRoutingTable from '../services/forwarding-routing-table'
 
 /**
  * Check whether a route can be filtered out based on DRAGON rules.
--- a/src/routing/dragon.ts
+++ b/src/routing/dragon.ts
@@ -3,6 +3,7 @@ import PrefixMap from './prefix-map'
 import { create as createLogger } from '../common/log'
 const log = createLogger('dragon')
 import { Relation, getRelationPriority } from './relation'
+import ForwardingRoutingTable from '../services/forwarding-routing-table';
 
 /**
  * Check whether a route can be filtered out based on DRAGON rules.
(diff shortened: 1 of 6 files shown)


src/routing/dragon.ts
```diff
@@ -3,6 +3,7 @@ import PrefixMap from './prefix-map'
 import { create as createLogger } from '../common/log'
 const log = createLogger('dragon')
 import { Relation, getRelationPriority } from './relation'
+import ForwardingRoutingTable from '../services/forwarding-routing-table';
```
sentientwaffle: Linting fails due to semicolon. Also, the `PrefixMap` import on line 2 isn't needed anymore.

test/route-broadcaster.test.js
```diff
@@ -231,6 +231,60 @@ describe('RouteBroadcaster', function () {
...
+        fromEpochIndex: 0,
+        toEpochIndex: 6,
+        holdDownTime: 45000,
+        withdrawnRoutes: [
+          'test.usd-ledger',
+          'test.cad-ledger'
+        ],
+        newRoutes: routesWithSourceLedgerB.filter(r => r.prefix !== 'test.cad-ledger')
+      }) || true))
+      throw new Error()
```
sentientwaffle: What is this `throw` for?

src/services/route-broadcaster.ts
```diff
@@ -406,11 +407,13 @@ export default class RouteBroadcaster {
...
-      routingLog: this.forwardingRoutingTable.log.map(entry => ({
-        ...entry,
-        route: entry.route && formatRouteAsJson(entry.route)
-      })),
+      forwardingRoutingTable: formatForwardingRoutingTableAsJson(this.forwardingRoutingTable),
+      routingLog: this.forwardingRoutingTable.log
+        .filter(Boolean)
+        .map(entry => ({
+          ...entry,
+          route: entry && entry.route && formatRouteAsJson(entry.route)
```
sentientwaffle: I think `entry && ` is unnecessary because of the preceding `.filter(Boolean)`.

src/services/route-broadcaster.ts
```diff
@@ -481,24 +484,29 @@ export default class RouteBroadcaster {
...
       const routeUpdate: RouteUpdate = {
         prefix,
         route,
         epoch
       }
+
+      if (route) {
+        this.forwardingRoutingTable.insert(prefix, routeUpdate)
+      } else {
+        this.forwardingRoutingTable.insert(prefix, routeUpdate)
```
sentientwaffle: Should this be `.delete`?

src/services/route-broadcaster.ts
justmoon: No, unfortunately, TypeScript isn't smart enough, so the `entry &&` is needed to make the type checker happy. :/

src/services/route-broadcaster.ts
justmoon: No, we need to keep track of the epoch of the last route change so we can later find it and remove it from the log easily. So we now want to keep an entry regardless of whether the route was added or removed. But, obviously, the `if` statement is superfluous now. D'oh. :)

PR #436, opened 2018-06-21
feat: admin-api: add alerting for unexpected T04

sentientwaffle:
Alerts are created when a peer unexpectedly returns a "maximum balance exceeded" error. They are accessible via the admin api. Also adds an endpoint to the admin API that allows the peer balances to be modified.

### DIFF
--- a/src/middlewares/alert.ts
+++ b/src/middlewares/alert.ts
@@ -1,0 +1,58 @@
+import { create as createLogger } from '../common/log'
+const log = createLogger('alert-middleware')
+import { Middleware, MiddlewareCallback, Pipelines } from '../types/middleware'
+import * as IlpPacket from 'ilp-packet'
+
+const { T04_INSUFFICIENT_LIQUIDITY } = IlpPacket.Errors.codes
+
+export interface Alert {
+  id: number
+  accountId: string
+  createdAt: Date
+  message: string
+}
+
+export default class AlertMiddleware implements Middleware {
+  private alerts: {[id: string]: Alert} = {}
+  private nextAlertId: number = Date.now()
+
+  async applyToPipelines (pipelines: Pipelines, accountId: string) {
+    pipelines.outgoingData.insertLast({
+      name: 'alert',
+      method: async (data: Buffer, next: MiddlewareCallback<Buffer, Buffer>) => {
+        const result = await next(data)
+        if (result[0] !== IlpPacket.Type.TYPE_ILP_REJECT) return result
+
+        const rejectPacket = IlpPacket.deserializeIlpReject(result)
+        if (rejectPacket.code !== T04_INSUFFICIENT_LIQUIDITY) return result
+
+        // The peer rejected a packet which, according to the local balance, should
+        // have succeeded. This can happen when our local connector owes the peer
+        // money but restarted before it was settled.
+        if (rejectPacket.message !== 'exceeded maximum balance.') return result
+
+        log.warn('generating alert for account=%s message="%s"', accountId, rejectPacket.message)
+
+        const id = this.nextAlertId++
+        this.alerts[id] = {
+          id,
+          accountId,
+          message: rejectPacket.message,
+          createdAt: new Date()
+        }
+
+        return result
+      }
+    })
+  }
+
+  getAlerts (): Alert[] {
+    return Object.keys(this.alerts)
+      .map((id) => this.alerts[id])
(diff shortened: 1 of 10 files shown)


src/services/admin-api.ts
```diff
@@ -72,62 +98,92 @@ export default class AdminApi {
...
+    const route = this.routes.find((route) =>
+      route.method === req.method && urlPrefix.startsWith(route.match))
+    if (!route) {
+      res.statusCode = 404
+      res.setHeader('Content-Type', 'text/plain')
+      res.end('Not Found')
+      return
+    }
+
+    const resBody = await route.fn.call(this, req.url, body && JSON.parse(body))
```
sharafian: If this throws will it automatically throw a 500 or will it hang?

src/middlewares/balance.ts
```diff
@@ -236,4 +211,48 @@ export default class BalanceMiddleware implements Middleware {
     })
     return { accounts }
   }
+
+  modifyBalance (accountId: string, _amountDiff: BigNumber.Value): BigNumber {
```
sharafian: I think it would be good to prominently log something in here (at warn level, probably), so that if you look through the logs you can see whether the balance is being unexpectedly changed. Otherwise if an attacker got access to your admin API they could modify the balance without leaving any trail

src/services/admin-api.ts
```diff
@@ -32,6 +46,18 @@ export default class AdminApi {
     this.routeBroadcaster = deps(RouteBroadcaster)
     this.rateBackend = deps(RateBackend)
     this.stats = deps(Stats)
+
+    this.routes = [
```
sharafian: With this many endpoints do you think it could be worth upgrading to something like koa?

src/services/admin-api.ts
sentientwaffle: `handleRequest` is called with a `catch` in `listen()` that will return a 500 on error.

src/middlewares/alert.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        // The peer rejected a packet which, according to the local balance, should
+        // have succeeded. This can happen when our local connector owes the peer
+        // money but restarted before it was settled.
+        if (rejectPacket.message !== 'exceeded maximum balance.') return result
+
+        log.warn('generating alert for account=%s message="%s"', accountId, rejectPacket.message)
+
+        const id = this.nextAlertId++
+        this.alerts[id] = {
```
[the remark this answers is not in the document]
sentientwaffle: What do you think of a `count` field on the alert? If an alert occurs, it will check if one with the same peer/message combination already exists, and if so, increment the count instead of creating a new alert.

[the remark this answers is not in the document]
sharafian: that sounds great, yeah

src/middlewares/alert.ts
```diff
@@ -0,0 +1,75 @@
...
+        const rejectPacket = IlpPacket.deserializeIlpReject(result)
+        if (rejectPacket.code !== T04_INSUFFICIENT_LIQUIDITY) return result
+
+        // The peer rejected a packet which, according to the local balance, should
+        // have succeeded. This can happen when our local connector owes the peer
+        // money but restarted before it was settled.
+        if (rejectPacket.message !== 'exceeded maximum balance.') return result
+
+        log.warn('generating alert for account=%s message="%s"', accountId, rejectPacket.message)
+        this.addAlert(accountId, rejectPacket.message)
```
sharafian: I think we should associate by the error's triggeredBy instead of by the accountId. There could be many routes that you reach through one peer, but different destinations on that route throw T04 errors

test/admin.test.js
```diff
@@ -285,4 +284,79 @@ describe('AdminApi', function () {
...
+      assert(false)
+    })
+  })
+
+  describe('getAlerts', function () {
+    it('returns no alerts by default', async function () {
+      assert.deepEqual(await this.adminApi.getAlerts(), {alerts: []})
+    })
+
+    it('returns an alert when a peer returns "maximum balance exceeded"', async function () {
```
sharafian: can we also add a test to make sure it correctly groups alerts which have been triggered by the same connector?

src/middlewares/alert.ts
sentientwaffle: I think I'll associate by the triple (accountId, triggeredBy, message)

PR #437, opened 2018-06-22
feat: trace level logging

njlie:


src/common/log.ts
```diff
@@ -1,15 +1,45 @@
...
+import * as debug from 'debug'
 
 import through2 = require('through2')
 const logStream = through2()
 logStream.pipe(process.stdout)
 
 // TODO: Not clear why I needed this, but got a
 //   TypeScript error without it.
-export interface ConnectorLogger extends riverpig.Logger { }
+export class ConnectorLogger {
```
sharafian: you could extend riverpig.Logger and add only the `trace` function

src/common/log.ts
```diff
@@ -1,15 +1,45 @@
...
+  error (...msg): void {
+    this.river.error(msg)
+  }
+
+  debug (...msg): void {
+    this.river.debug(msg)
+  }
+
+  trace (...msg): void {
+    this.tracer(msg)
```
sharafian: shouldn't it be `this.tracer(...msg)`?

src/common/log.ts
njlie: The riverpig.Logger  that is exported is a type alias, rather than a class. It doesn't appear to export its own Logger class. I think extending this will still require me to define a class that fits the extended type definition

src/common/log.ts
```diff
@@ -1,15 +1,44 @@
...
+
+  error (msg: any, ...elements: any[]): void {
+    this.river.error(msg, ...elements)
+  }
+
+  debug (msg: any, ...elements: any[]): void {
+    this.river.debug(msg, ...elements)
+  }
+
+  trace (msg: any, ...elements: any[]): void {
```
sharafian: why require two arguments instead of doing just `(...msg: any[])`?

njlie: Typescript is enforcing that the function calls here match the function signatures for their interfaces. Both debug and riverpig implement their logging functions like this, so they expect at least one argument and using just the rest parameter will cause it to complain that it got at least 0 args on compile.

src/common/log.ts
```diff
@@ -1,15 +1,44 @@
 import * as riverpig from 'riverpig'
+import { LoggerConfig } from 'riverpig'
+import { Logger } from 'riverpig'
```
traviscrist: Couldn't this be one statement, `import { LoggerConfig, Logger } from 'riverpig'` or am I missing something here?

src/common/log.ts
traviscrist: Should we add a comment about this for others who might have the same question?

PR #442, opened 2018-07-01
feat: use Prometheus-compatible metrics

adrianhopebailie:
BREAKING: Admin API `/stats` endpoint uses new format

This change modifies the admin API as follows:

1. The `/stats` endpoint is a new more verbose format to better reflect the new data model for metrics.
2. There is a new `/metrics` endpoint for consumption by a Prometheus scraper job allowing the data to be recorded as time series and properly analyzed.

### DIFF
--- a/src/middlewares/balance.ts
+++ b/src/middlewares/balance.ts
@@ -180,13 +180,13 @@ export default class BalanceMiddleware implements Middleware {
             try {
               result = await next(data)
             } catch (err) {
-              log.debug('outgoing packet refunded due to ilp reject. accountId=%s amount=%s newBalance=%s', accountId, parsedPacket.amount, balance.getValue())
+              log.debug('outgoing packet not applied due to error. accountId=%s amount=%s newBalance=%s', accountId, parsedPacket.amount, balance.getValue())
               this.stats.outgoingDataPacketValue.increment(account, { result : 'failed' }, +parsedPacket.amount)
               throw err
             }
 
             if (result[0] === IlpPacket.Type.TYPE_ILP_REJECT) {
-              log.debug('outgoing packet refunded due to ilp reject. accountId=%s amount=%s newBalance=%s', accountId, parsedPacket.amount, balance.getValue())
+              log.debug('outgoing packet not applied due to ilp reject. accountId=%s amount=%s newBalance=%s', accountId, parsedPacket.amount, balance.getValue())
               this.stats.outgoingDataPacketValue.increment(account, { result : 'rejected' }, +parsedPacket.amount)
             } else if (result[0] === IlpPacket.Type.TYPE_ILP_FULFILL) {
               // Decrease balance on prepare
--- a/src/middlewares/balance.ts
+++ b/src/middlewares/balance.ts
@@ -174,30 +174,26 @@ export default class BalanceMiddleware implements Middleware {
               return next(data)
             }
 
-            balance.subtract(parsedPacket.amount)
-            log.debug('balance decreased due to outgoing ilp prepare. accountId=%s amount=%s newBalance=%s', accountId, parsedPacket.amount, balance.getValue())
-            this.stats.balance.setValue(account, {}, balance.getValue().toNumber())
-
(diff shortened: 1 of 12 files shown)


src/middlewares/balance.ts
```diff
@@ -162,25 +174,31 @@ export default class BalanceMiddleware implements Middleware {
               return next(data)
             }
 
+            balance.subtract(parsedPacket.amount)
```
adrianhopebailie: @justmoon the logic for incoming and outgoing packets was previously different. Was there a reason for this?

I assume the thinking is that awe want to be greedy on incoming and conservative on outgoing packets however this seems we may be at risk of forwarding packets we can afford because we think our balance is higher than it is.

Specifically, incoming packets cause the balance to be increased on prepare and then decreased if failed or rejected

Previously, outgoing packets did not impact the balance until the fulfil was received. I have changed the logic to be the same as for incoming.

I can revert this part of the change or put in a different PR.

justmoon: Yeah, the worry is that someone can prepare a bunch of outgoing packets (by sending them from anywhere else in the network such that we're in the path and they are the next hop after us), triggering a settlement and then cancel them and run away with the money. So we don't want to settle outgoing payments until we receive a fulfillment.

You're right that there is an asymmetry, i.e. you'll subtract my balance with you (on prepare) before I increase your balance with me (on fulfill) but I believe that this is correct. In a peering relationship like this, I have my view of the balance and you have yours. I should update my view in a way that is conservative from my perspective and you should update yours in a way that is conservative from your perspective.

> we may be at risk of forwarding packets we can afford because we think our balance is higher than it is.

This problem should be addressed with configuration. Our balances can differ at most by the max in-flight amount. So my min balance should be larger than your max by at least the max in-flight amount.

Your change could work, too. Note that it increases the max at risk from `max balance` to `max balance + max in-flight amount`. Not setting a max in-flight amount would now expose you to unlimited risk. But I could see your way being closer to what people expect.

The downside is that it is closer to what people expect due to a misunderstanding: That there is only one balance between us. That's not true. I have a balance and you have a balance and those can arbitrarily differ. Nothing we can do about that. So my instinct was to make that difference more explicit rather than to minimize it.

I think it would be more likely that we (the developers) or the users make a mistake that is exploitable with this change than without it. With the current way, if you configure it naively the risk is that packets get rejected. With the proposed new way, if you configure it naively the risk is that money is stolen from you.

adrianhopebailie: Agreed. Have reverted those changes.

src/services/admin-api.ts
```diff
@@ -14,16 +14,17 @@ import { formatRoutingTableAsJson } from '../routing/utils'
...
+import * as Prometheus from 'prom-client'
 const log = createLogger('admin-api')
 const ajv = new Ajv()
 const validateBalanceUpdate = ajv.compile(require('../schemas/BalanceUpdate.json'))
 
 interface Route {
   method: 'GET' | 'POST' | 'DELETE'
   match: string
-  fn: (url: string, body: object) => Promise<object | void>
+  fn: (url: string, body: Object) => Promise<Object | string | void>
```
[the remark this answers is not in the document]
adrianhopebailie: Left over from some experiments with returning alt content types. Reverted.

src/services/admin-api.ts
```diff
@@ -144,7 +151,7 @@ export default class AdminApi {
     return balanceMiddleware.getStatus()
   }
 
-  private async postBalance (url: string, _data: object) {
+  private async postBalance (url: string, _data: Object) {
```
sentientwaffle: `s/object/Object`

src/services/stats.ts
```diff
@@ -1,20 +1,84 @@
...
 
   getStatus () {
-    return {
-      counters: this.counters,
-      meters: this.meters
-    }
+    return Prometheus.register.getMetricsAsJSON()
   }
+
+  collectDefaultMetrics () {
```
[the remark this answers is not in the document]
adrianhopebailie: Not anymore. I previously called it by default when the Admin API was enabled but we can add these if required.

PR #460, opened 2018-08-08
Feat: add admin info interface to accounts

sharafian:
Right now, the kinds of things that the Moneyd GUI can view at are pretty limited. We can only see the internals of the connector.

This feature allows us to look inside of the plugin and allow it to expose additional info and admin operations. It adds two optional functions to the plugin interface.

```ts
getAdminInfo (): Promise<object>
setAdminInfo (object): Promise<object> 
```

It also adds two endpoints to the connector's admin API.

```http
GET /accounts/:account_id
POST /accounts/:account_id
```

`getAdminInfo` resolves an object with information about the plugin's state. For an XRP plugin, for instance, it might return channel IDs and channel balances, along with overall XRP balance. You can get this information from the admin API. Moneyd GUI could display a nicely formatted interface for plugins it knows about, and could dump the raw JSON for plugins it doesn't recognize.

`setAdminInfo` allows a JSON object to be passed into the plugin, and comes back with a result. For an XRP plugin, you could use this to manually trigger a channel fund or a settlement to a child (in the case of asym-server). It would make server plugins in general a lot less opaque. Moneyd GUI could display forms and buttons for plugins it knows about, or allow users to write raw JSON to send to plugins that Moneyd GUI doesn't recognize.

This change is non-breaking, but adds fields to existing admin API responses.

### DIFF
--- a/src/services/accounts.ts
+++ b/src/services/accounts.ts
@@ -216,7 +216,8 @@ export default class Accounts extends EventEmitter {
       accounts[accountId] = {
         // Set info.options to undefined so that credentials aren't exposed.
         info: Object.assign({}, account.info, { options: undefined }),
-        connected: account.plugin.isConnected()
+        connected: account.plugin.isConnected(),
+        adminInfo: !!account.plugin.getAdminInfo
       }
     })
     return {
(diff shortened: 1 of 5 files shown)


sharafian:
Maybe `postAdminInfo`?

sharafian:
changed to `sendAdminInfo` after discussion in order to be most consistent with `sendData` and `sendMoney`

PR #465, opened 2018-10-26
 fixes for connector.addPlugin/connector.removePlugin

sentientwaffle:
- Fix startup pipeline setup.
- Add a `teardown` pipeline to correctly handle plugins being removed with `app.removePlugin()`. Right now the only use is in the deduplicate middleware to disable the `setInterval`.
- Allow `app.addPlugin` to pass a plugin instance.
- Ensure that the `startup` pipeline is run for plugins added after the connector is already running.
- call `reloadLocalRoutes()` when a plugin is added or removed so that it can be routed to.
- Support import via TypeScript w/ types


src/schemas/Config.json
```diff
@@ -35,8 +35,8 @@
             "enum": ["parent", "peer", "child"]
           },
           "plugin": {
-            "description": "Name of the ILP plugin that should be used for this account.",
-            "type": "string"
+            "description": "Name or instance of the ILP plugin that should be used for this account.",
```
justmoon: Might be worth clarifying that you can only pass an instance when instantiating the connector from JavaScript. If I'm reading the docs from the perspective of environment variables I might get a bit confused trying to pass objects here.

PR #466, opened 2018-10-31
fix: stats memory leak

sentientwaffle:
Labelling metrics with `triggeredBy` causes a memory leak because there are an unbounded number of ILP addresses that may be saved in memory.

Also fixes a typo: `s/AccountGuage/AccountGauge`.


sharafian:
Does this stop our T04 detection code from working? Or is the alerts page separate?

sentientwaffle:
Alerts are handled separately, this patch doesn't affect them.

PR #469, opened 2018-11-13
feat: standalone coinmarketcap backend

karzak:


src/backends/coinmarketcap.ts
```diff
@@ -0,0 +1,88 @@
...
+const RATES_API = 'https://api.coinmarketcap.com/v2/ticker/'
+
+export default class CMCBackend implements BackendInstance {
+  public rates: {}
+  protected spread: number
+  protected ratesApiUrl: string
+  protected getInfo: (accountId: any) => AccountInfo | undefined
+
+  constructor (opts: CMCBackendOpts, api: BackendServices) {
+    this.spread = opts.spread | 0
```
sentientwaffle: `|` is a bitwise OR (which might actually work in this case since `undefined | 0 === 0`), but it should probably use a logical OR `||` instead for clarity.

karzak:
Going to work on something more permanent.

PR #470, opened 2018-11-26
chore: reenable integration tests

matdehaast:
Re enable the integration tests now that the integration tests have been moved to use ILPv4 with STREAM

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "lint": "tslint --project . && eslint test/",
     "test": "nyc mocha",
     "report-coverage": "nyc report --reporter=json && codecov -f coverage/*.json",
-    "integration": "integration-loader && integration all",
+    "integration": "integration-loader && integration all -- --exit",
     "commitmsg": "commitlint -e $GIT_PARAMS"
   },
   "engines": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "lint": "tslint --project . && eslint test/",
     "test": "nyc mocha",
     "report-coverage": "nyc report --reporter=json && codecov -f coverage/*.json",
-    "integration": "integration-loader && integration all",
+    "integration": "integration-loader && integration all -- --exit",
```
sentientwaffle: What is `--exit`? I don't see it in https://github.com/interledgerjs/five-bells-integration-test/blob/master/src/bin/integration.

matdehaast: As there are unterminated listeners lingering in tests/tested components (STREAM and possible Connector), Mocha 4 doesnt automatically exit the tests (https://github.com/mochajs/mocha/issues/3044)

Adding the `-- --exit` passes the exit flag to mocha and allows the tests to exit correctly after finished running

sentientwaffle: Unterminated listeners seems like a bug that should be fixed, not sidestepped, but this is an alright stopgap.

PR #479, opened 2019-03-01
feat: broaden scope of backend service

tangsauce:
- `controllers/ilp-prepare` - change `submitPayment` to `submitPacket`: always submit to backend whether packet type is fulfill, prepare, reject
- `types/backend` - add more params to `SubmitPaymentParams`: backend should receive more data
- `services/backend` - replace `this.getInfo` with `this.accounts`: backend should have access to more of the `accounts` api
- backends `ecb`, `one-to-one`, `randomizer`: update from using `this.getInfo` to `this.accounts`

### DIFF
--- a/src/controllers/ilp-prepare.ts
+++ b/src/controllers/ilp-prepare.ts
@@ -47,6 +47,18 @@ export default class IlpPrepareController {
     log.trace('sending outbound ilp prepare. destination=%s amount=%s', destination, nextHopPacket.amount)
     const result = await outbound(IlpPacket.serializeIlpPrepare(nextHopPacket), nextHop)
 
+    this.backend.submitPacket({
+      sourceAccount: sourceAccount,
+      sourceAmount: amount,
+      destinationAccount: nextHop,
+      destinationAmount: nextHopPacket.amount,
+      parsedPacket,
+      result
+    }).catch(err => {
+      const errInfo = (err && typeof err === 'object' && err.stack) ? err.stack : String(err)
+      log.error('error while submitting packet to backend. error=%s', errInfo)
+    })
+
     if (result[0] === IlpPacket.Type.TYPE_ILP_FULFILL) {
       log.trace('got fulfillment. cond=%s nextHop=%s amount=%s', executionCondition.slice(0, 6).toString('base64'), nextHop, nextHopPacket.amount)
 
--- a/src/services/rate-backend.ts
+++ b/src/services/rate-backend.ts
@@ -23,7 +23,8 @@ export default class RateBackend implements BackendInstance {
     this.backend = new Backend(Object.assign({
       spread: config.spread
     }, config.backendConfig), {
-      getInfo: (account: string) => this.accounts.getInfo(account)
+      getInfo: (account: string) => this.accounts.getInfo(account),
+      accounts: this.accounts
     })
   }
 
@@ -37,6 +38,13 @@ export default class RateBackend implements BackendInstance {
 
   submitPayment (params: SubmitPaymentParams) {
     return this.backend.submitPayment(params)
+  }
+
+  submitPacket (params: SubmitPaymentParams) {
+    if (this.backend.submitPacket) {
+      return this.backend.submitPacket(params)
+    }
+    return Promise.resolve()
   }
 
   async getStatus () {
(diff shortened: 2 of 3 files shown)


sharafian:
There are a lot of breaking changes here that could be done non-breakingly. These are good features though, especially because they're informed by real-world usage of the connector.

- Instead of replacing `getInfo` with `accounts` you could add `accounts` and still keep `getInfo`
- `submitPacket` could run on prepare/fulfill/reject, and on fulfill if `submitPacket` is not specified it can fall back to `submitPayment`

tangsauce:
This might be a dumb question, but where are the breaking changes? I thought I modified the backends (ecb, one-to-one, randomizer) such that they use `submitPacket` and `this.accounts.getInfo` rather than `submitPayment` and `this.getInfo`, respectively.

kincaidoneil:
For example, we wrote our own [custom backend](https://github.com/Kava-Labs/ilp-backend-crypto). Since it imports and implements the backend interface from the connector, this change would break it since it relies on `getInfo`, and would introduce type errors.

tangsauce:
With help from @sharafian and @kincaidoneil :

- in `types/backend`: `accounts` property optional, `getInfo` method unchanged
- in `types/backend`: `submitPacket` method optional, `submitPayment` method unchanged
- in `controllers/ilp-prepare`: call `submitPacket` on prepare/fulfill/reject if method exists, otherwise call `submitPayment` on fulfill
- in `services/rate-backend`: on construction, assign `this.submitPayment` to be `this.backend.submitPayment`, otherwise, leave it undefined

Hopefully these changes improve the extensibility of custom backends, while maintaining backwards-compatibility.

tangsauce:
Awesome - can @sentientwaffle  also take a look at these two PRs? They follow the paradigm you initially suggested of passing in a server object to the respective constructors.

1. https://github.com/interledgerjs/ilp-plugin-btp/pull/53
2. https://github.com/interledgerjs/ilp-plugin-mini-accounts/pull/43/files

src/controllers/ilp-prepare.ts
```diff
@@ -47,19 +47,35 @@ export default class IlpPrepareController {
...
       })
-        .catch(err => {
-          const errInfo = (err && typeof err === 'object' && err.stack) ? err.stack : String(err)
-          log.error('error while submitting payment to backend. error=%s', errInfo)
+    }
+
+    if (result[0] === IlpPacket.Type.TYPE_ILP_FULFILL) {
+      log.trace('got fulfillment. cond=%s nextHop=%s amount=%s', executionCondition.slice(0, 6).toString('base64'), nextHop, nextHopPacket.amount)
+
+      if (!this.backend.submitPacket) {
```
sentientwaffle: Wouldn't `if (this.backend.submitPayment)` be clearer as a condition?

src/services/rate-backend.ts
```diff
@@ -12,6 +12,7 @@ import { loadModuleOfType } from '../lib/utils'
 const DEFAULT_BACKEND = 'one-to-one'
 
 export default class RateBackend implements BackendInstance {
+  public submitPacket: any
```
sentientwaffle: Use the callback's fn signature here.

src/services/rate-backend.ts
```diff
@@ -23,8 +24,12 @@ export default class RateBackend implements BackendInstance {
     this.backend = new Backend(Object.assign({
       spread: config.spread
     }, config.backendConfig), {
-      getInfo: (account: string) => this.accounts.getInfo(account)
+      getInfo: (account: string) => this.accounts.getInfo(account),
+      accounts: this.accounts
     })
+    if (this.backend.submitPacket) {
```
sentientwaffle: Instead of conditionally attaching this function and then having callers have to test for it, why not add a method to `RateBackend` that does `submitPacket(...) { if (this.backend.submitPacket) { this.backend.submitPacket(...) }}`. `submitPayemnt` could do the same.

src/controllers/ilp-prepare.ts
tangsauce: It would be if `submitPayment` was optional, but it is required and that's not something I wanted to change.

src/services/rate-backend.ts
tangsauce: Resolved by comment:

> Instead of conditionally attaching this function and then having callers have to test for it, why not add a method to RateBackend that does submitPacket(...) { if (this.backend.submitPacket) { this.backend.submitPacket(...) }}. submitPayemnt could do the same.

src/services/rate-backend.ts
tangsauce: Great idea!

PR #482, opened 2019-03-08
Update typescript version

sharafian:
Closes https://github.com/interledgerjs/ilp-connector/issues/480

This fixes errors when importing ILP Connector to the latest version of typescript. Also makes the type checking stricter by enabling `noImplicitAny`

This doesn't break anything if you're still importing from typescript 2.6

### DIFF
--- a/src/services/admin-api.ts
+++ b/src/services/admin-api.ts
@@ -23,7 +23,7 @@ const validateBalanceUpdate = ajv.compile(require('../schemas/BalanceUpdate.json
 interface Route {
   method: 'GET' | 'POST' | 'DELETE'
   match: string
-  fn: (url: string, body: object) => Promise<object | string | void>
+  fn: (url: string | undefined, body: object) => Promise<object | string | void>
   responseType?: string
 }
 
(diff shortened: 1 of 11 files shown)


src/stores/leveldown.ts
```diff
@@ -23,9 +23,14 @@ class LeveldownStore implements StoreInstance {
     this.db = levelup(leveldown(path))
   }
 
-  async get (key: string) {
+  async get (key: string): Promise<string | undefined> {
```
sentientwaffle: Why does this have `| undefined`?

src/services/admin-api.ts
```diff
@@ -153,7 +153,7 @@ export default class AdminApi {
     return balanceMiddleware.getStatus()
   }
 
-  private async postBalance (url: string, _data: object) {
+  private async postBalance (url: string | undefined, _data: object) {
```
sentientwaffle: Why `| undefined`?

sharafian: `req.url` is `string | undefined` now, so the url params need to become `string | undefined`

src/stores/leveldown.ts
sharafian: if the key doesn't exist then it returns `undefined`

PR #486, opened 2019-04-05
chore: update to circleci 2

sentientwaffle:
@d1no007: this is why ilp-connector didn't publish properly before


.circleci/config.yml
```diff
@@ -0,0 +1,60 @@
...
+            branches:
+              ignore:
+                - gh-pages
+      - publish_to_npm:
+          requires:
+            - build
+          filters:
+            branches:
+              only:
+                - master
```
adrianhopebailie: Do you want to limit this to only tagged releases?

PR #489, opened 2019-06-10
Merge Prometheus ILP metrics with default metrics

martinlowinski:
As discussed on Slack, this is a PR to merge the metrics from the Prometheus ILP registry with the default registry and export it via the endpoint `/metrics`.

Prior to v22.2.0, the metrics export via `/metrics` used the Prometheus default registry that contains both metrics (ILP and nodejs/process) while >22.2.0 creates its own registry for the ILP metrics. This additional registry is only exported as JSON via the `/stats` endpoint, but not via `/metrics` anymore.

### DIFF
--- a/src/services/admin-api.ts
+++ b/src/services/admin-api.ts
@@ -199,7 +199,10 @@ export default class AdminApi {
   }
 
   private async getMetrics () {
-    return Prometheus.register.metrics()
+    const promRegistry = Prometheus.register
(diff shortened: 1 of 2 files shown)


adrianhopebailie:
LGTM although @sentientwaffle may have some comments regarding things he's seen related to performance in the past.

Would like to get his input before merging.

PR #491, opened 2019-07-16
Added openapi spec under documentation directory

AyAggarwal:
v.2, being developed here: https://studio.apicur.io/apis/12550


martinlowinski:
Maybe it makes sense to add the `.yaml` extension to the spec file?

adrianhopebailie:
I cleaned up the .DS files and renamed the file to match common patterns (and have an extension)

sappenin (changes requested):
I think we should add a `README.md` to the `documentation` directory to outline what's in here. Some things that would be helpful:

1. Overview of the `documentation` folder (what should and should not go in here, what might be coming in the future).
1. Instructions/suggestions  for how to view the OpenAPI in a UI (e.g., probably just use [SwaggerHub Editor](https://editor.swagger.io/))

Also, there's already _some_ documentation in the project's [README](https://github.com/interledgerjs/ilp-connector/blob/master/README.md), so reserving this folder by itself feels a bit off. Perhaps we should rename this folder to be `/documentation/api-specs`. This would leave room for future "documentation" in the root `/documentation` folder.

AyAggarwal:
@sappenin I added a README, and reformatted the directory. My thought is that the Reference section from the root directory's readme could go in this documentation directory.

sappenin (approved):
LGTM!

martinlowinski (approved):

westurner:
Is there another source for ILP specs as OpenAPI specs?

PR #495, opened 2020-01-16
feat: BREAKING no more leveldown

sharafian:
The leveldown module packaged with the connector is unused in almost all cases and causes a plethora of build errors and other issues when installing the connector. This PR removes the leveldown store and the memdown store and replaces them with a memory store that is functionally the same as the memdown one but doesn't require a dependency.

This is a breaking change. If there's interest in keeping the leveldown store around it could be moved into a standalone module but AFAIK nobody is still using it. Even the balance middleware is typically not used in production so on most deployments no store is being used.


kincaidoneil:
> The leveldown module packaged with the connector is unused in almost all cases and causes a plethora of build errors and other issues when installing the connector

👍 

> Even the balance middleware is typically not used in production so on most deployments no store is being used.

FYI, I think the balance middleware was already entirely in-memory... https://github.com/interledgerjs/ilp-connector/issues/478

PR #496, opened 2020-05-05
fix: only publish compiled typescript

kincaidoneil:
Since `ilp-connector` published its TS source to NPM, I've run into TS build errors in modules that depend on it if they use a stricter/incompatible tsconfig. Now, only the type declarations and transpiled JS will be published.

Also, includes a fix for an issue with the exported types, an issue that prevented shutdown, and some other minor clean-up.


kincaidoneil:
Superceded by #497 -- CI doesn't seem to like forks

PR #497, opened 2020-05-05
fix: only publish transpiled typescript

kincaidoneil:
Since `ilp-connector` published its TS source to NPM, I've run into TS build errors in modules that depend on it if they use a stricter/incompatible tsconfig. Now, only the type declarations and transpiled JS will be published.

Also, includes a fix for an issue with the exported types, an issue that prevented shutdown, and some other minor clean-up.


.circleci/config.yml
```diff
@@ -6,55 +6,56 @@ jobs:
...
   version: 2
   build_and_publish:
     jobs:
       - build:
           filters:
-            branches:
-              ignore:
-                - gh-pages
+            tags:
+              only: /.*/
```
sentientwaffle, line 53: What is this filter for?

test/admin.test.js
```diff
@@ -6,7 +6,7 @@ const { cloneDeep } = require('lodash')
 const IlpPacket = require('ilp-packet')
 const appHelper = require('./helpers/app')
 const logHelper = require('./helpers/log')
-const logger = require('../src/common/log')
+const logger = require('../dist/common/log')
```
sentientwaffle, line 9: Why have the tests `require` modules from `dist` instead of `src`? (In general, not this file in particular).

.circleci/config.yml
kincaidoneil: This was copied from the second example [here](https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag). Since the `publish_to_npm` job has tag filters and requires the `build` job, the `build` job must also match any tags the `publish_to_npm` job matches -- so I believe the filter matches any tag

test/admin.test.js
kincaidoneil: The transpiled JS is only in `dist/` now, so unless the tests were upgraded to TypeScript they can't require anything in `src/`

package.json
```diff
@@ -8,28 +8,28 @@
...
+    "start-prof": "node --prof --logfile=${CONNECTOR_V8_LOGFILE:-v8.log} dist/index.js",
+    "start:watch": "nodemon dist/index.js",
+    "build": "npm run schema-to-tsd && npm run compile-ts && npm run copy-schemas",
     "schema-to-tsd": "node scripts/build_schema_types",
     "compile-ts": "tsc --project .",
+    "copy-schemas": "cpy 'src/schemas/*.json' dist/schemas",
     "prepack": "npm run build",
-    "prepare": "npm run build",
     "lint": "tslint --project . && eslint test/*.test.js test/mocks/ test/helpers/",
     "test": "nyc mocha",
```
sentientwaffle, line 34: Since `npm test` now tests `dist/` not `src/`, what do you think of adding:

```suggestion
    "test": "npm build && nyc mocha",
```

to make sure the `npm test` doesn't get out of sync?

PR #498, opened 2020-09-21
fix(echo): apply business rules

kincaidoneil:
Echo "pong" packets were forwarded without a cross-currency conversion to the destination asset, expiry checking, and incorrect deduction of the min message window.

This changes the `echo` implementation to a middleware so all necessary business rules are applied.

Also:
- Add tests for ping/pong packets
- Remove some legacy ILP tests
- Fix package vulnerabilities
- Remove CoinMarketCap backend (API is deprecated/non-functional)
- Minor cleanup to README


sentientwaffle (commented):
Add the `"echo"` middleware to the `disableMiddleware` items list in `Config.json` and the list of middlewares in the `README.md`.

sentientwaffle (approved):
LGTM

